PR #36, opened 2021-01-19
Fix typo in extents array constructor

mattja:
Fixes the constructor that takes dynamic extents as a std::array.


r-burns:
The compilation errors from this had me scratching my head... thanks for the fix :sweat_smile:

dalg24 (approved):

PR #38, opened 2021-01-25
Add test coverage for stdex::extents

mattja:
Test coverage for extents constructors and observers.
Tested for C++11, C++14, C++17, C++20 with gcc 10.2.

### DIFF
--- a/tests/test_extents.cpp
+++ b/tests/test_extents.cpp
@@ -1,0 +1,176 @@
+/*
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 2.0
+//              Copyright (2021) Sandia Corporation
+//
+// Under the terms of Contract DE-AC04-94AL85000 with Sandia Corporation,
+// the U.S. Government retains certain rights in this software.
+//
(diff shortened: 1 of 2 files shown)


tests/test_extents.cpp
```diff
@@ -0,0 +1,176 @@
...
+  EXPECT_EQ(e, this->exts);
+}
+
+TYPED_TEST(TestExtents, copy_ctor) {
+  typename TestFixture::extents_type e { this->exts };
+  EXPECT_EQ(e, this->exts);
+}
+
+TYPED_TEST(TestExtents, copy_assign) {
+  typename TestFixture::extents_type e = this->exts;
```
dalg24, line 122: This is not testing the copy assignment operator.  It uses the copy constructor.

dalg24 (approved):

tests/test_extents.cpp
dalg24: Will fix in a follow up PR

mattja: Thanks for this correction - just saw it.

PR #43, opened 2021-05-11
Fix span access using a std::array of indices

tpadioleau:
This PR attemps to solve an access issue in `basic_mdspan` when using a `std::array` of indices such as

```c++
void foo(std::experimental::mdspan<int, 2, 2> sdata) {
  std::array<ptrdiff_t, 2> multi_index{0, 0};
  sdata(multi_index) = 4; // This does not compile
}
```


dalg24 (approved):

dalg24:
Didn't mean to close...

PR #47, opened 2021-05-16
add an implicit accessor conversion operator and a test

### LINKED ISSUE #44: Assigning to a mdspan to const
The snippet below does not seem to compile
```c++
std::vector<double> v(10);
stdex::mdspan<double, stdex::dynamic_extent> a(v.data(), v.size());
stdex::mdspan<double const, stdex::dynamic_extent> b = a; // This currently does not compile
```

The compilation error seems to come from the `accessor_basic` that is not convertible from `accessor_basic<double>` to `accessor_basic<double const>`.

I think one might expect to be able to assign a `mdspan` to a `mdspan`  to `const` as it is the case for `std::span`.

A fix could be to add a conversion operator like this one
```c++
MDSPAN_INLINE_FUNCTION
constexpr operator accessor_basic<std::add_const_t<ElementType>>() const noexcept {
  return {};
}
```

tpadioleau:
Attempt to fix #44

### DIFF
--- a/include/experimental/__p0009_bits/accessor_basic.hpp
+++ b/include/experimental/__p0009_bits/accessor_basic.hpp
@@ -46,6 +46,7 @@
 #include "macros.hpp"
 
 #include <cstddef> // ptrdiff_t
+#include <type_traits> // add_const
 
 namespace std {
 namespace experimental {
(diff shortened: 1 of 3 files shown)


crtrott:
Should we do this as a converting constructor instead?

dalg24:
> Should we do this as a converting constructor instead?

See https://github.com/kokkos/mdspan/issues/44#issuecomment-842411872

nliber:
I agree that it should be done as a converting constructor (and we've updated P0009R12 to that effect).

An issue with the way it is currently implemented (unconstrained conversion operator) is that a if we have an `mdspan<const double, 1>` a compiler may warn about this (both clang and gcc do in the non-template class case), because it is a conversion to itself.

tpadioleau:
I guess this PR can be closed then. I will let you do the renaming and write this new constructor.

PR #48, opened 2021-05-18
Enable github actions

amklinv-nnl:


dalg24:
Looks ok to me.  Is there a way to run to make sure it works before we merge?

masterleinad:
> Looks ok to me. Is there a way to run to make sure it works before we merge?

https://github.com/amklinv-nnl/mdspan/runs/2614298747?check_suite_focus=true

dalg24 (approved):

PR #49, opened 2021-05-19
Fix non compiling code

keryell:
Probably some old refactoring which went untested, breaking code like:

  const std::array<std::size_t, 2> index { 1, 2 };
  some_mdspan(index);


dalg24 (changes requested):
This looks like a duplicate of #46

keryell:
Indeed! How did I miss this? It looks like I started diving into this before https://github.com/kokkos/mdspan/pull/46 landed. :-(
Actually https://github.com/kokkos/mdspan/pull/46 seems to be closer to the intent from @dhollman and there are even some tests!
So, closing this.

PR #50, opened 2021-05-21
Add converting constructor to basic_accessor

dalg24:
Alternate solution to #44 
I cherry-picked the test added in #47 and added a converting constructor to `accessor_basic` per our discussion.

### DIFF
--- a/include/experimental/__p0009_bits/accessor_basic.hpp
+++ b/include/experimental/__p0009_bits/accessor_basic.hpp
@@ -52,11 +52,22 @@ namespace experimental {
 
 template <class ElementType>
 struct accessor_basic {
-  
+
   using offset_policy = accessor_basic;
   using element_type = ElementType;
   using reference = ElementType&;
   using pointer = ElementType*;
+
+  constexpr accessor_basic() noexcept = default;
+
+  MDSPAN_TEMPLATE_REQUIRES(
+    class OtherElementType,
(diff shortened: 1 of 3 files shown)


include/experimental/__p0009_bits/accessor_basic.hpp
```diff
@@ -52,12 +52,23 @@ namespace experimental {
...
+  constexpr accessor_basic() noexcept = default;
+
+  MDSPAN_TEMPLATE_REQUIRES(
+    class OtherElementType,
+    /* requires */ (
+      _MDSPAN_TRAIT(is_convertible, typename accessor_basic<OtherElementType>::pointer, pointer)
+    )
+  )
+  MDSPAN_INLINE_FUNCTION
+  constexpr accessor_basic(accessor_basic<OtherElementType>) {}
```
dalg24, line 70: @nliber should we declare it `explicit`?

crtrott (approved):

include/experimental/__p0009_bits/accessor_basic.hpp
crtrott: this way we don't need assign explicitly. But if Nevin thinks its better.

nliber: I think implicit is correct (and it almost never matters for templated constructors).

PR #57, opened 2021-06-08
Fixup missing return statement in predicate in `layout_stride::is_contiguous()`

### LINKED ISSUE #55: Compile error when using `is_contiguous()` on subspan
Using `is_contiguous()` on a subspan will result in a compile error. Consider following example:
```

std::array<int, 3 * 3> buf;

const stdex::mdspan<int, stdex::dynamic_extent, stdex::dynamic_extent> span(buf.data(),3,3);
const auto subspan = stdex::subspan(span, std::pair(1, 2), std::pair(1, 2));

subspan.is_contiguous(); // will cause build error
```
The relevant lines in the list of traces msvc is spitting out are the following:
```
D:\...\mdspan\include\experimental\__p0009_bits\layout_stride.hpp(245): warning C4553: '==': result of expression not used; did you intend '='?
C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\VC\Tools\MSVC\14.29.30129\include\xutility(5637): error C2451: a conditional expression of type 'void' is not valid
```
The error is caused by a missing return statement in a predicate in the `is_contiguous()` implementation in `layout_stride.hpp`:
```
auto next_idx_iter = std::find_if(
  rem.begin(), rem.end(),
  [&](ptrdiff_t i) { this->stride(i) == 1;  }
  //                ^-- missing return
);
```

dalg24:
Fix #55

### DIFF
--- a/tests/test_layout_stride.cpp
+++ b/tests/test_layout_stride.cpp
@@ -119,6 +119,7 @@ TEST(TestLayoutStrideListInitialization, test_list_initialization) {
   ASSERT_EQ(m.stride(1), 128);
   ASSERT_EQ(m.extent(0), 16);
   ASSERT_EQ(m.extent(1), 32);
+  ASSERT_FALSE(m.is_contiguous());
 }
 
 // GCC 10 ICEs if I stick this in the body of the below test.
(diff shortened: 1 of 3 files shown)


tests/test_layout_stride.cpp
```diff
@@ -119,6 +119,7 @@ TEST(TestLayoutStrideListInitialization, test_list_initialization) {
   ASSERT_EQ(m.stride(1), 128);
   ASSERT_EQ(m.extent(0), 16);
   ASSERT_EQ(m.extent(1), 32);
+  ASSERT_FALSE(m.is_contiguous());
```
dalg24, line 122: This assertion triggers the bug that was reported.
The others are drive-by changes

crtrott (approved):

PR #60, opened 2021-06-10
Fix gcc 5,6 and intel 17,18 in C++14

crtrott:
Also removed some C++11 leftover config stuff.


dsunder (approved):
Looks good though I don't have access to an Intel compiler to check.

nliber (approved):

PR #61, opened 2021-07-02
Support using system-installed GTest

r-burns:
Adds CMake option for using a local GoogleTest installation when downloading at build time is not desired/allowed.

My personal motivation for this feature is to support running the unit tests while packaging for Nix, which doesn't allow networking at build time. I think a fair number of other package managers also prohibit networking at build time, although I don't know how common it is to try to run upstream tests.


mhoemmen:
We certainly do maintain mdspan.  Perhaps we missed this PR.  Let me take a look.

mhoemmen (approved):
Thanks for the contribution!

As a future improvement, we might want to consider always doing `find_package(GTest CONFIG)`, and only attempting the download if that fails.

mhoemmen (approved):

PR #71, opened 2021-10-12
Fixes missing LayoutStride conversion constructor and extent != comparison

crtrott:


crtrott:
These things showed up during Kokkos integration, for now I am piggy bagging on Kokkos testing for this. When the inegration of mdspan into Kokkos is fully done, we will set up CI testing here which builds a fixed Kokkos version against the proposed mdspan change.

mhoemmen (approved):
A+

dalg24:
Where is the unit test that makes sure I don't break tomorrow?

Issue #63, opened 2021-08-05
Submdspan issue


tpadioleau:
I am trying the `submdspan` function and I have unexpected results. Here is my test

```cpp
#include <array>
#include <cstdint>
#include <experimental/mdspan>
#include <iostream>

using namespace std::experimental;

template <class ET, class EP, class LP, class AP>
void print_layout(mdspan<ET, EP, LP, AP> const &t) {
  if constexpr (std::is_same_v<LP, layout_left>) {
    std::cout << "layout=left";
  } else if constexpr (std::is_same_v<LP, layout_right>) {
    std::cout << "layout=right";
  } else if constexpr (std::is_same_v<LP, layout_stride>) {
    std::cout << "layout=stride";
  }
}

int main(int argc, char **argv) {
  std::array<double, 100> raw{};
  mdspan<double, dextents<2>, layout_right> s(raw.data(), 10, 10);

  std::cout << "Expected: layout=right, strides=[10,1], extents=[10,10]\n";
  std::cout << "Result: ";
  print_layout(s);
  std::cout << ", strides=[" << s.stride(0) << "," << s.stride(1) << "]";
  std::cout << ", extents=[" << s.extent(0) << "," << s.extent(1) << "]" << std::endl;
  std::cout << std::endl;

  auto s1 = submdspan(s, full_extent, std::pair(3ul, 7ul));
  std::cout << "Expected: layout=stride, strides=[10,1], extents=[10,4]\n";
  std::cout << "Result: ";
  print_layout(s1);
  std::cout << ", strides=[" << s1.stride(0) << "," << s1.stride(1) << "]";
  std::cout << ", extents=[" << s1.extent(0) << "," << s1.extent(1) << "]" << std::endl;

  std::cout << std::endl;
  auto s2 = submdspan(s, std::pair(3ul, 7ul), full_extent);
  std::cout << "Expected: layout=right, strides=[10,1], extents=[4,10]\n";
  std::cout << "Result: ";
  print_layout(s2);
  std::cout << ", strides=[" << s2.stride(0) << "," << s2.stride(1) << "]";
  std::cout << ", extents=[" << s2.extent(0) << "," << s2.extent(1) << "]" << std::endl;

  return 0;
}
```

The output is the following where I print my expectation and the result:

```
Expected: layout=right, strides=[10,1], extents=[10,10]
Result: layout=right, strides=[10,1], extents=[10,10]

Expected: layout=stride, strides=[10,1], extents=[10,4]
[...]

PR #72, opened 2021-10-12
Fix layout preservation in submdspan

jbigot:
As identified in issue #63 , the logic for preserving layout left/right was incorrect. This fixes it.


crtrott:
Ah let me test this, but I think this may fix one of my issues :-)

crtrott (approved):

crtrott:
Thanks for this. I wrote tests for this which live on a different branch. Will merge those after #88.

PR #76, opened 2021-10-27
Implement missing `strides` function of `layout_stride::mapping`

tpadioleau:
I propose an implementation for the missing `strides` member function with some tests.


mhoemmen (approved):
Thanks for the fix! : - )

PR #77, opened 2021-11-10
Adopt new extents explicit rules

crtrott:
This changes the extents constructors to be explicit for construction from integers and when converting runtime to compile time extents

### DIFF
--- a/include/experimental/__p0009_bits/config.hpp
+++ b/include/experimental/__p0009_bits/config.hpp
@@ -232,8 +232,10 @@ static_assert(_MDSPAN_CPLUSPLUS >= MDSPAN_CXX_STD_14, "mdspan requires C++14 or 
 #  endif
 #endif
 
-#ifndef _MDSPAN_USE_CONDITIONAL_EXPLICIT
+#ifndef MDSPAN_CONDITIONAL_EXPLICIT
 #  if (defined(MDSPAN_HAS_CXX_20))
-#    define _MDSPAN_USE_CONDITIONAL_EXPLICIT 1
+#    define MDSPAN_CONDITIONAL_EXPLICIT(COND) explicit(COND)
+#  else
+#    define MDSPAN_CONDITIONAL_EXPLICIT(COND)
 #  endif
 #endif
--- a/include/experimental/__p0009_bits/config.hpp
+++ b/include/experimental/__p0009_bits/config.hpp
@@ -231,3 +231,9 @@ static_assert(_MDSPAN_CPLUSPLUS >= MDSPAN_CXX_STD_14, "mdspan requires C++14 or 
 #    endif
 #  endif
 #endif
+
+#ifndef _MDSPAN_USE_CONDITIONAL_EXPLICIT
+#  if (defined(MDSPAN_HAS_CXX_20))
+#    define _MDSPAN_USE_CONDITIONAL_EXPLICIT 1
+#  endif
+#endif
--- a/include/experimental/__p0009_bits/layout_left.hpp
+++ b/include/experimental/__p0009_bits/layout_left.hpp
@@ -110,9 +110,12 @@ struct layout_left {
     MDSPAN_TEMPLATE_REQUIRES(
       class OtherExtents,
       /* requires */ (
-        _MDSPAN_TRAIT(is_convertible, OtherExtents, Extents)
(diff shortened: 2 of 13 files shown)


crtrott:
Note this has a number of downstream consequences. Some of the constructors which worked before with initializer lists don't work anymore. Not sure if that is fixeable, but I don't believe so since they rely on implicit construction of extents from just integers.

crtrott:
Note I commented a few construction tests out instead of deleting them so you can see also what used to work.

include/experimental/__p0009_bits/config.hpp
```diff
@@ -231,3 +231,9 @@ static_assert(_MDSPAN_CPLUSPLUS >= MDSPAN_CXX_STD_14, "mdspan requires C++14 or
 #    endif
 #  endif
 #endif
+
+#ifndef _MDSPAN_USE_CONDITIONAL_EXPLICIT
+#  if (defined(MDSPAN_HAS_CXX_20))
+#    define _MDSPAN_USE_CONDITIONAL_EXPLICIT 1
+#  endif
```
dalg24:
```suggestion
#  if (defined(MDSPAN_HAS_CXX_20))
#    define _MDSPAN_CONDITIONAL_EXPLICIT(COND) explicit(COND)
#  else
#    define _MDSPAN_CONDITIONAL_EXPLICIT(COND)
#  endif
```

include/experimental/__p0009_bits/layout_left.hpp
```diff
@@ -110,9 +110,12 @@ struct layout_left {
     MDSPAN_TEMPLATE_REQUIRES(
       class OtherExtents,
       /* requires */ (
-        _MDSPAN_TRAIT(is_convertible, OtherExtents, Extents)
+        _MDSPAN_TRAIT(is_constructible, OtherExtents, Extents)
       )
     )
+#ifdef _MDSPAN_USE_CONDITIONAL_EXPLICIT
+    explicit (!is_convertible<OtherExtents, Extents>::value)
+#endif
```
dalg24:
```suggestion
    _MDSPAN_CONDITIONAL_EXPLICIT (!is_convertible<OtherExtents, Extents>::value)
```

dalg24 (changes requested):

include/experimental/__p0009_bits/config.hpp
mhoemmen: That is nicer, yes : - )

include/experimental/__p0009_bits/layout_left.hpp
mhoemmen: Would the comma inside a macro invocation need protection (by wrapping the whole expression in an extra pair of parens)?

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -214,11 +215,12 @@ struct layout_stride {
     // TODO @proposal-bug In the proposal, the constructor doesn't take a
```
mhoemmen: Would this TODO still be valid after this change?

include/experimental/__p0009_bits/mdspan.hpp
```diff
@@ -143,14 +143,14 @@ class mdspan
   )
   MDSPAN_INLINE_FUNCTION
   // TODO @proposal-bug Why is this explicit?
```
mhoemmen: Did we fix these TODOs in the proposal?  Had thought so.

include/experimental/__p0009_bits/layout_left.hpp
dalg24: Probably so

include/experimental/__p0009_bits/layout_stride.hpp
crtrott: yeah that is still an issue in the proposal. Lets fix it there.

crtrott: https://github.com/ORNL/cpp-proposals-pub/pull/169

include/experimental/__p0009_bits/mdspan.hpp
crtrott: yeah deleted the comments.

dalg24 (approved):

PR #80, opened 2021-11-15
Allow extents construction from all extents

crtrott:
This addresses one of the changes in P0009R14: we should allow constructions of extents (and mdspan - but I haven't done that yet) from either just `dynamic_rank` values or `rank` values. Both for integer packs and array.

### DIFF
--- a/include/experimental/__p0009_bits/extents.hpp
+++ b/include/experimental/__p0009_bits/extents.hpp
@@ -216,18 +216,21 @@ public:
     class... Integral,
     /* requires */ (
       _MDSPAN_FOLD_AND(_MDSPAN_TRAIT(is_convertible, Integral, size_type) /* && ... */)
-        && sizeof...(Integral) == rank_dynamic()
+        && ((sizeof...(Integral) == rank_dynamic()) ||
+            (sizeof...(Integral) == rank()))
     )
   )
   MDSPAN_INLINE_FUNCTION
-  explicit constexpr extents(Integral... dyn) noexcept
+  explicit constexpr extents(Integral... exts) noexcept
 #if defined(_MDSPAN_USE_ATTRIBUTE_NO_UNIQUE_ADDRESS)
     : __storage_{
 #else
     : __base_t(__base_t{typename __base_t::__stored_type{
 #endif
-        detail::__construct_partially_static_array_from_sizes_tag,
-        detail::__construct_partially_static_array_from_sizes_tag, static_cast<size_t>(dyn)...
+      std::conditional_t<sizeof...(Integral)==rank_dynamic(),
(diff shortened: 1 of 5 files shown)


include/experimental/__p0009_bits/extents.hpp
```diff
@@ -237,20 +240,24 @@ class extents
...
-  extents(std::array<SizeType, rank_dynamic()> const& dyn) noexcept
+  extents(std::array<SizeType, N> const& exts) noexcept
 #if defined(_MDSPAN_USE_ATTRIBUTE_NO_UNIQUE_ADDRESS)
     : __storage_{
 #else
     : __base_t(__base_t{typename __base_t::__stored_type{
 #endif
-        detail::__construct_psa_from_dynamic_values_tag_t<>{}, dyn
+      std::conditional_t<N==rank_dynamic(),
+        detail::__construct_psa_from_dynamic_values_tag_t<0>,
```
mhoemmen: Would you consider spelling out "psa" as we already do in `__construct_partially_static_array_from_{all,dynamic}_sizes_tag_t`?  That would improve consistency.

tests/test_extents.cpp
```diff
@@ -151,12 +151,59 @@ struct TestExtentsCompatCtors<std::tuple<
...
+  template<class Exts, size_t Idx>
+  struct make_extents<Exts,Idx,false,true>
+  {
+    template<class ... SizeTypes>
+    static Exts construct(SizeTypes...sizes) {
+      return make_extents<Exts, Idx-1,false,(Idx>1?Exts::static_extent(Idx-2)!=stdex::dynamic_extent:true)>::construct(sizes...);
+    }
+  };
+
+  // End case
```
mhoemmen, line 186: Is it ever possible for `Idx` to be zero and `Static` to be `false`?  If so, then would you consider adding a base case for `Static=false` as well?

mhoemmen (dismissed):

tests/test_extents.cpp
crtrott: no its not. The actual iteration with something happening is 1<=Idx<=Rank. I make sure that for Idx zero its always Static=true.

include/experimental/__p0009_bits/extents.hpp
crtrott: I went the other way and now have:
```
__construct_psa_from_all_exts_values_tag_t
__construct_psa_from_dynamic_exts_values_tag_t
__construct_psa_from_all_exts_array_tag_t
__construct_psa_from_dynamic_exts_array_tag_t
```
For the four code paths.

crtrott: writing this out was just getting too long

PR #85, opened 2021-11-16
Layout Left right conversion and tests. 

crtrott:
Part of P0009R14, only allows conversion of Rank 0 and Rank 1.

### DIFF
--- a/tests/test_contiguous_layouts.cpp
+++ b/tests/test_contiguous_layouts.cpp
@@ -180,6 +180,7 @@ using test_layout_conversion =
 
 using layout_conversion_test_types =
   ::testing::Types<
+    // Left <=> Left conversion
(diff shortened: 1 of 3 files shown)


tests/test_contiguous_layouts.cpp
```diff
@@ -166,3 +166,96 @@ TYPED_TEST(TestLayout3D, mapping_works) {
 TYPED_TEST(TestLayout3D, required_span_size_works) {
   ASSERT_EQ(this->map.required_span_size(), 3*4*5);
 }
+
+template<bool val>
+struct bool_wrapper {
```
mhoemmen, line 171: Would you consider just using `std::integral_constant<bool, val>`?

mhoemmen (approved):

tests/test_contiguous_layouts.cpp
crtrott: sure I fix that in the follow on I am working on.

PR #88, opened 2021-11-17
Add operator bracket and makes test work with them 

crtrott:
Three kinda configure macros but not really (i.e. they are not CMake options, but you can override the autodetection) are included:
MDSPAN_USE_BRACKET_OPERATOR=[0,1] // enables bracket operator
MDSPAN_USE_PAREN_OPERATOR=[0,1] // enables paren operator
MDSPAN_NO_EMPTY_BRACKET_OPERATOR // disables use of [] without arguments

I do believe the latter is just not implemented in corentins compiler, but that we should support that ...

### DIFF
--- a/compilation_tests/ctest_constexpr_submdspan.cpp
+++ b/compilation_tests/ctest_constexpr_submdspan.cpp
@@ -370,12 +370,12 @@ submdspan_single_element_stress_test_impl_2(
   auto ss_dyn = stdex::submdspan(s_dyn, _repeated_ptrdiff_t<0, Idxs>...);
   auto ss_all = stdex::submdspan(s, _repeated_with_idxs_t<stdex::full_extent_t, Idxs>{}...);
   auto ss_all_dyn = stdex::submdspan(s_dyn, _repeated_with_idxs_t<stdex::full_extent_t, Idxs>{}...);
-  auto val = __MDSPAN_OP(ss_all, _repeated_ptrdiff_t<0, Idxs>...);
-  auto val_dyn = __MDSPAN_OP(ss_all_dyn, _repeated_ptrdiff_t<0, Idxs>...);
+  auto val = __MDSPAN_OP(ss_all, (_repeated_ptrdiff_t<0, Idxs>...));
+  auto val_dyn = __MDSPAN_OP(ss_all_dyn, (_repeated_ptrdiff_t<0, Idxs>...));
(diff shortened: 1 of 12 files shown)


compilation_tests/ctest_constexpr_submdspan.cpp
```diff
@@ -370,19 +370,19 @@ submdspan_single_element_stress_test_impl_2(
...
   auto ss_all_dyn = stdex::submdspan(s_dyn, _repeated_with_idxs_t<stdex::full_extent_t, Idxs>{}...);
-  auto val = ss_all(_repeated_ptrdiff_t<0, Idxs>...);
-  auto val_dyn = ss_all_dyn(_repeated_ptrdiff_t<0, Idxs>...);
+  auto val = __MDSPAN_OP(ss_all, _repeated_ptrdiff_t<0, Idxs>...);
+  auto val_dyn = __MDSPAN_OP(ss_all_dyn, _repeated_ptrdiff_t<0, Idxs>...);
   auto ss_pair = stdex::submdspan(s, _repeated_with_idxs_t<std::pair<ptrdiff_t, ptrdiff_t>, Idxs>{0, 1}...);
   auto ss_pair_dyn = stdex::submdspan(s_dyn, _repeated_with_idxs_t<std::pair<ptrdiff_t, ptrdiff_t>, Idxs>{0, 1}...);
-  auto val_pair = ss_pair(_repeated_ptrdiff_t<0, Idxs>...);
-  auto val_pair_dyn = ss_pair_dyn(_repeated_ptrdiff_t<0, Idxs>...);
+  auto val_pair = __MDSPAN_OP(ss_pair, _repeated_ptrdiff_t<0, Idxs>...);
```
mhoemmen: Does the second macro argument need to be protected by parentheses, since it has a comma?

crtrott: hm doesn't seem like it. The compiler complained in a number of other places but not for that pattern ?

crtrott: maybe its because of the parameter pack expansion?

crtrott: actually its probably because the macro itself is variadic. but basically I don't know what the right thing is, it also works with the parenthesis around it. I just add them for now and we can revisit if it breaks somewhere. After all its only a test.

crtrott:
Merging since clang is broken without it ...

crtrott:
if you have more change requests please comment more here and I will work those in tomorrow.

crtrott:
@dalg24 @mhoemmen

PR #101, opened 2021-11-29
Change `mdspan::rank[_dynamic]` return type `{int -> size_t}`

### LINKED ISSUE #99: mdspan.hpp (line 268): conversion from size_t to int
Calling rank() gives a conversion from size_t to int, possible loss of data warning at line 268 in mdspan.hpp

dalg24:
Fix #99
We already had `extents<Extents...>::rank[_dynamic] -> size_t` and this appear to be an oversight on our end.


mhoemmen (approved):
Thanks for the fix! : - )

PR #106, opened 2021-12-09
Compile all tests as CUDA code + test for extents on device

crtrott:


include/experimental/__p0009_bits/extents.hpp
```diff
@@ -242,8 +263,16 @@ class extents
   MDSPAN_TEMPLATE_REQUIRES(
     class SizeType, size_t N,
     /* requires */ (
-      (N == rank() || N == rank_dynamic()) &&
       _MDSPAN_TRAIT(is_convertible, SizeType, size_type)
+// TODO: check whether this works with newest NVCC, doesn't with 11.4
```
nmm0, line 267: Does `rank()` and `rank_dynamic()` not work at all with NVCC or just in the REQUIRES or FOLD_AND macros?

include/experimental/__p0009_bits/layout_left.hpp
```diff
@@ -143,12 +143,14 @@ struct layout_left {
     mapping(OtherMapping const& other) // NOLINT(google-explicit-constructor)
       :__extents(other.extents())
     {
+       #ifndef __CUDA_ARCH__
```
nmm0, line 146: Is it worth having some checking on this in a kernel, with the kernel aborting if the strides don't match?

tests/offload_utils.hpp
```diff
@@ -0,0 +1,59 @@
+
+namespace {
```
nmm0, line 2: fine that this is anonymous unless we have multiple test translation units using this

nmm0 (dismissed):
Looks good to me -- just a few nonblocking issues/questions about the code

include/experimental/__p0009_bits/extents.hpp
crtrott: in most cases its correct, just in that fold expression as part of some funky enable ifing it falls apart.

include/experimental/__p0009_bits/layout_left.hpp
crtrott: yeah eventually we can do this, but I think as a first step its fine not to.

tests/offload_utils.hpp
crtrott: we do, but the point is that it needs to be local to each translation unit.

crtrott:
Note the primary reason that I went with the rather crude global flag method for testing both device and host, is that I wanted to avoid writing a mini-Kokkos with templating on execution spaces etc. I think this is good enough, and least overhead (in terms of maintenance/development cost) way of achieving this goal.

nmm0:
Apparently my approval doesn't count because I don't have write access to the repo :P

compilation_tests/CMakeLists.txt
```diff
@@ -9,8 +12,17 @@ add_compilation_test(ctest_extents_ctors)
...
+if(MDSPAN_ENABLE_CUDA)
+  # For NVCC 11.5 and older we do not do no_unique_address optimization
+  # thus all these strict size tests here will fail
+  if(CMAKE_CUDA_COMPILER_VERSION VERSION_GREATER "11.5")
+    add_compilation_test(ctest_no_unique_address)
+    add_compilation_test(ctest_compressed_pair_layout)
+  endif()
+else()
 add_compilation_test(ctest_no_unique_address)
 add_compilation_test(ctest_compressed_pair_layout)
```
dalg24: Please indent

include/experimental/__p0009_bits/config.hpp
```diff
@@ -124,6 +126,15 @@ static_assert(_MDSPAN_CPLUSPLUS >= MDSPAN_CXX_STD_14, "mdspan requires C++14 or
 #  endif
 #endif
 
+// NVCC older than 11.6 chokes on the no-unique-address-emulation
+// so just pretend to use it, and leave the macro empty
```
dalg24: Why do we pretend we are using it?

include/experimental/__p0009_bits/layout_left.hpp
```diff
@@ -143,12 +143,14 @@ struct layout_left {
...
        size_t stride = 1;
        for(size_type r=0; r<__extents.rank(); r++) {
          if(stride != other.stride(r))
            throw std::runtime_error("Assigning layout_stride to layout_left with invalid strides.");
```
dalg24, line 150: Any reason not to just do
```C++
assert(stride == other.stride(r) && "Assigning layout_stride to layout_left with invalid strides.");

File: include/experimental/__p0009_bits/layout_stride.hpp [hunk 7]
```diff
@@ -309,7 +311,7 @@ struct layout_stride {
         size_t prev_stride_times_prev_extent =
           this->extents().extent(*next_idx_iter) * this->stride(*next_idx_iter);
         // "remove" the index
-        constexpr size_t removed_index_sentinel = -1;
+        constexpr size_t removed_index_sentinel = size_t(-1);
```
dalg24: What was wrong before?
```suggestion
        constexpr auto removed_index_sentinel = static_cast<size_t>(-1);
```

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -298,7 +298,9 @@ struct layout_stride {
     MDSPAN_INLINE_FUNCTION constexpr bool is_unique() const noexcept { return true; }
     MDSPAN_INLINE_FUNCTION _MDSPAN_CONSTEXPR_14 bool is_contiguous() const noexcept {
       // TODO @testing test layout_stride is_contiguous()
-
+#ifdef __CUDA_ARCH__
+      return false;
```
dalg24, line 303: Is it a FIXME?

include/experimental/__p0009_bits/standard_layout_static_array.hpp
```diff
@@ -539,7 +539,9 @@ struct __partially_static_sizes_tagged
   using __psa_impl_t = __standard_layout_psa<
       _Tag, size_t, integer_sequence<size_t, __values_or_sentinals...>>;
   using __psa_impl_t::__psa_impl_t;
+#ifdef _MDSPAN_DEFAULTED_CONSTRUCTORS_INHERITANCE_WORKAROUND
```
dalg24, line 542: This is not what this macro is for

tests/offload_utils.hpp
```diff
@@ -0,0 +1,59 @@
...
+__global__ void dispatch_kernel(const LAMBDA f) {
+  f();
+}
+
+template<class LAMBDA>
+void dispatch(const LAMBDA f) {
+  if(dispatch_host) {
+    f();
+  } else {
+    dispatch_kernel<<<1,1>>>(f);
```
dalg24:
```suggestion
void dispatch(LAMBDA&& f) {
  if(dispatch_host) {
    (LAMBDA&&)f();
  } else {
    dispatch_kernel<<<1,1>>>((LAMBDA&&)f);
```

tests/offload_utils.hpp
```diff
@@ -0,0 +1,59 @@
...
+    cudaFree(ptr);
+}
+
+#define __MDSPAN_TESTS_DISPATCH_DEFINED
+#endif
+
+#ifndef __MDSPAN_TESTS_DISPATCH_DEFINED
+template<class LAMBDA>
+void dispatch(const LAMBDA f) {
+  f();
```
dalg24:
```suggestion
void dispatch(LAMBDA&& f) {
  (LAMBDA&&)f();
```

tests/test_extents.cpp
```diff
@@ -42,23 +42,86 @@
...
 
 #include <experimental/mdspan>
-
+#include "offload_utils.hpp"
 #include <gtest/gtest.h>
 
 namespace stdex = std::experimental;
 
+// Making actual test implementations part of the class,
+// can't create CUDA lambdas in the primary test fucntions, since they are private
```
dalg24:
```suggestion
// can't create CUDA lambdas in the primary test functions, since they are private
```

tests/test_extents.cpp
```diff
@@ -277,6 +339,7 @@ struct ImplicitConversionToExts<Extents,false> {
 
 
 TYPED_TEST(TestExtentsCompatCtors, implicit_construct_1) {
+#ifdef _MDSPAN_USE_CONDITIONAL_EXPLICIT
```
dalg24, line 332: Then disable the test altogether
```C++
#ifdef _MDSPAN_USE_CONDITIONAL_EXPLICIT
TYPED_TEST(...

File: tests/test_extents.cpp [hunk 14]
```diff
@@ -79,25 +142,24 @@ using extents_test_types =
...
 TYPED_TEST(TestExtents, rank) {
-  EXPECT_EQ(this->exts.rank(), this->static_sizes.size());
-}
-
-TYPED_TEST(TestExtents, dynamic_rank) {
-  EXPECT_EQ(this->exts.rank_dynamic(), this->dyn_sizes.size());
+  dispatch_host = true;
+  this->test_rank();
+  dispatch_host = false;
+  this->test_rank();
```
dalg24: Meaning we run twice w/o CUDA.
If we always have that pattern, can't we encapsulate it?

include/experimental/__p0009_bits/config.hpp
crtrott: because otherwise the code try to uses a complicated EBO workaround which fails horribly with NVCC

include/experimental/__p0009_bits/layout_left.hpp
crtrott: no

include/experimental/__p0009_bits/layout_stride.hpp
crtrott: some warning

include/experimental/__p0009_bits/config.hpp
dalg24: OK please comment to that effect

tests/offload_utils.hpp
dalg24: Nevermind it doesn't like the member functions of your fixture class

compilation_tests/CMakeLists.txt
crtrott: done

include/experimental/__p0009_bits/config.hpp
crtrott: done

include/experimental/__p0009_bits/layout_left.hpp
crtrott: actually:
```
Assertions should only be used to verify conditions that should be logically impossible to be false (read: sanity checks). These conditions should only be based on inputs generated by your own code. Any checks based on external inputs should use exceptions.
```

crtrott: https://docs.microsoft.com/en-us/cpp/cpp/errors-and-exception-handling-modern-cpp?view=msvc-170

include/experimental/__p0009_bits/layout_stride.hpp
crtrott: yes

dalg24 (approved):

include/experimental/__p0009_bits/standard_layout_static_array.hpp
dalg24: Ping

dalg24: My bad

PR #110, opened 2021-12-20
Make an mdspan unit test compile on GPU

crtrott:


tests/test_mdspan_ctors.cpp
```diff
@@ -45,20 +45,37 @@
...
 namespace stdex = std::experimental;
 _MDSPAN_INLINE_VARIABLE constexpr auto dyn = stdex::dynamic_extent;
 
+
+void test_mdspan_ctor_data_carray() {
+  size_t* result = allocate_array<size_t>(6);
+  dispatch([=] _MDSPAN_HOST_DEVICE () {
+    int data[1] = {42};
+    stdex::mdspan<int, stdex::extents<1>> m(data);
+    result[0] = m.data() == data ? 1 : 0;
```
dalg24:
```C++
#define POOR_MAN_ASSERT_EQ(LHS, RHS) \
if (!(LHS == RHS)) {
  printf("expected equality of %s and %s", #LHS, #RHS); // can improve to print values when possible
  ++errors; // increment some error counter
}

Comment (mhoemmen) [reply tests/test_mdspan_ctors.cpp]: @dalg24 I have a vague memory that the Sierra (Sandia's finite-element suite) developers had written something like this.

Comment (crtrott) [reply tests/test_mdspan_ctors.cpp]: addressed

File: tests/offload_utils.hpp [hunk 2]
```diff
@@ -1,6 +1,13 @@
 
 namespace {
 bool dispatch_host = true;
+
+#define __MDSPAN_DEVICE_ASSERT_EQ(LHS, RHS) \
+if (!(LHS == RHS)) { \
+  printf("expected equality of %s and %s", #LHS, #RHS); \
```
dalg24:
```suggestion
  printf("expected equality of %s and %s\n", #LHS, #RHS); \
```

tests/offload_utils.hpp
```diff
@@ -1,6 +1,13 @@
 
 namespace {
 bool dispatch_host = true;
+
+#define __MDSPAN_DEVICE_ASSERT_EQ(LHS, RHS) \
```
dalg24, line 5: Double underscore prefix doesn't sound like a good idea.  I don't think it is legal.

dalg24 (dismissed):
Make sure you add the `'\n'` and you can merge as far as I am concerned

tests/offload_utils.hpp
crtrott: OK but then we need to rename a few others too. We can do that in a separate PR.

dalg24 (approved):

PR #118, opened 2022-01-09
Fix typo in godbolt starter example

tobiashienzsch:


mhoemmen (approved):
Thanks for the fix! : - )

PR #122, opened 2022-02-18
Implement p2554

crtrott:


include/experimental/__p0009_bits/mdspan.hpp
```diff
@@ -307,13 +307,38 @@ class mdspan
 };
 
 #if defined(_MDSPAN_USE_CLASS_TEMPLATE_ARGUMENT_DEDUCTION)
+#ifdef _MDSPAN_USE_P2554
```
dalg24, line 310: Why prefix with an underscore?

tests/test_mdspan_ctors.cpp
```diff
@@ -150,7 +150,49 @@ TEST(TestMdspanCTADExtentsPack, test_mdspan_ctad_extents_pack) {
...
-TEST(TestMdspanCTADExtentsObject, test_mdspan_ctad_extents_object) {
+TEST(TestMdspanCTAD, ctad_carray) {
+  int data[5] = {1,2,3,4,5};
+  stdex::mdspan m(data);
+  static_assert(std::is_same<decltype(m)::element_type,int>::value);
+  ASSERT_EQ(m.data(), &data[0]);
+  #ifdef  _MDSPAN_USE_P2554
+  ASSERT_EQ(m.rank(), 1);
+  ASSERT_EQ(m.rank_dynamic(), 0);
+  ASSERT_EQ(m.static_extent(0), 5);
```
dalg24, line 171: Consider using `static_assert` instead

tests/test_mdspan_ctors.cpp
```diff
@@ -150,7 +150,49 @@ TEST(TestMdspanCTADExtentsPack, test_mdspan_ctad_extents_pack) {
...
+  ASSERT_EQ(m.extent(0), 5);
+  #else
+  ASSERT_EQ(m.rank(), 0);
+  ASSERT_EQ(m.rank_dynamic(), 0);
+  #endif
+  ASSERT_TRUE(m.is_contiguous());
```
dalg24, line 178: Consider adding one assertion (static?) `__MDSPAN_OP(m, 1) == 2` for instance

include/experimental/__p0009_bits/mdspan.hpp
```diff
@@ -307,13 +307,38 @@ class mdspan
...
+  (sizeof...(SizeTypes) > 0)
+)
+mdspan(ElementType*, SizeTypes...)
+  -> mdspan<ElementType, ::std::experimental::dextents<sizeof...(SizeTypes)>>;
+
+MDSPAN_TEMPLATE_REQUIRES(
+  class Pointer,
+  (!_MDSPAN_TRAIT(is_array, Pointer))
+)
+mdspan(Pointer&) -> mdspan<std::remove_pointer_t<Pointer>, extents<>>;
```
dalg24: Add a test for that guy

include/experimental/__p0009_bits/mdspan.hpp
crtrott: we did that for all internal parameters. Its not as if I want downstream users to use this really. And when it gets approved I will remove the thing.

tests/test_mdspan_ctors.cpp
crtrott: I kept it in line with the other tests. We can certainly change that, but then I rather would want to change it consistently.

include/experimental/__p0009_bits/mdspan.hpp
dalg24: https://github.com/kokkos/mdspan/blob/82281329d69cfc88cfa1f13c47118795464e86cb/include/experimental/__p0009_bits/config.hpp#L255-L269

include/experimental/__p0009_bits/mdspan.hpp
crtrott: ok. Funnily enough I found a bug in the proposal through me writing the test slightly differently then what I had tested before. The deduction guide for this guy needs to take const Pointer& instead of Pointer& otherwise mdspan a(d.data()) doesn't work

include/experimental/__p0009_bits/mdspan.hpp
crtrott: that one is more of a public option. Since we do intend to publicly support both bracket and paren with this impl.

dalg24 (dismissed):

README.md
```diff
@@ -34,7 +34,7 @@ TODO write this
 Caveats
 -------
 
-This implementation is fully conforming with a few exceptions (most of which are extensions):
+This implementation is fully conforming with revision 14 of P0009 with a few exceptions (most of which are extensions):
```
mhoemmen, line 37: Wouldn't this be R15 now?

mhoemmen (dismissed):

README.md
crtrott: no because we are not conforming with it yet. I haven't gone through it yet to make all the changes we did with Tomasz.

dalg24 (approved):

PR #128, opened 2022-04-08
Add mdarray

crtrott:


include/experimental/__p1684_bits/mdarray.hpp
```diff
@@ -0,0 +1,484 @@
...
+  MDSPAN_INLINE_FUNCTION
+  constexpr mdarray(const mdarray<OtherElementType, OtherExtents, OtherLayoutPolicy, OtherContainer>& other)
+    : map_(other.mapping()), ctr_(other.container())
+  {
+    static_assert(is_constructible<extents_type, OtherExtents>::value, "");
+  }
+
+  // Constructors for container types constructible from a size and allocator
+  MDSPAN_TEMPLATE_REQUIRES(
+    class Alloc,
```
AnabelSMRuggiero, line 252: As is, I suspect these constructors are under constrained. Constructing an `mdarray<int, ..., std::vector<int>>` with an `extents_type` and a `float` (as an example), would satisfy the requirements of this constructor. That causes the resulting vector to have the appropriate number of elements, each initialized the the value of the (truncated) `float`. Not the worst thing to possibly happen here, but I've not exhaustively checked all constructors.

AnabelSMRuggiero (approved):
Other than my comment about under-constrained, but not invariant violating allocator constructors, nothing else seems out of line with the current proposal to me.

rscohn2:
Where can I see a spec for the proposal?

rscohn2:
> Where can I see a spec for the proposal?

Found it: http://www7.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p1684r1.html

include/experimental/__p1684_bits/mdarray.hpp
AnabelSMRuggiero: It occurred to me today `requires uses_allocator_v<container_type, Alloc>` or equivalent for pre-C++20 should be sufficient for constraining the allocator constructors.

mhoemmen:
@rscohn2 One can use the redirect link https://wg21.link/p1684 to find the latest published version of a proposal.

mhoemmen (approved):
Approving because this is already an improvement; we can come back to other issues later : - )

Excellent test using PMR btw!

PR #129, opened 2022-04-26
Expands MDSPAN_TEMPLATE_REQUIRES to support templates with more template parameters

youyu3:
So that stdBLAS can use this instead of directly using `requires` for backwards-compatibility.


mhoemmen (approved):
Thanks for this change!  I think this is fine for now.  In the future, we may want to consider rewriting this to use `...` and `__VA_ARGS__`, if possible.  That would call for changing the interface to put the requirement first, like this:

```c++
#define MDSPAN_TEMPLATE_REQUIRES_N(REQ, ...) \
  template<__VA_ARGS__ __VA_OPT__(,) \
    MDSPAN_CLOSE_ANGLE_REQUIRES(REQ)
```

I think it's OK to wait on that, though, because it would require touching a lot of mdspan and downstream code.

crtrott (approved):

Issue #131, opened 2022-05-23
Compilation with non-gnu compilers and `-pedantic`


chrisrichardson:
When compiling with e.g. clang and `-pedantic` the following type of warnings are triggered:

```
mdspan.hpp:574:9: warning: token pasting of ',' and __VA_ARGS__ is a GNU extension [-Wgnu-zero-variadic-macro-arguments]
```
For some applications, we are using `-Werror -pedantic` which then causes this code to fail...

crtrott:
see #132

crtrott:
Was on vacation a while, closing this assuming your thumbsup meant that worked for you.

PR #132, opened 2022-05-23
Fix warning about variable macro args with no args

crtrott:
Turns out we have to go through the somewhat more complex
MSVC variant of counting args, in order to avoid the warning.

Addresses #131

### DIFF
--- a/tests/test_mdarray_ctors.cpp
+++ b/tests/test_mdarray_ctors.cpp
@@ -117,8 +117,10 @@ struct mdarray_values<2> {
   template<class MDA>
   static void check(const MDA& m) {
     for(int i=0; i<m.extent(0); i++)
-      for(int j=0; j<m.extent(1); j++)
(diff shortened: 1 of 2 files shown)


tests/test_mdarray_ctors.cpp
```diff
@@ -117,8 +117,10 @@ struct mdarray_values<2> {
   template<class MDA>
   static void check(const MDA& m) {
     for(int i=0; i<m.extent(0); i++)
-      for(int j=0; j<m.extent(1); j++)
-        ASSERT_EQ(__MDSPAN_OP(m,i,j), 42 + i*1000 + j);
+      for(int j=0; j<m.extent(1); j++) {
+        auto tmp = __MDSPAN_OP(m,i,j);
+        ASSERT_EQ(tmp, 42 + i*1000 + j);
+      }
```
dalg24, line 123: Why did you change this?

crtrott: nested macro madness.

dalg24 (approved):

PR #135, opened 2022-06-09
Skip architecture check on the installed package

rafal-c:
Skip architecture check on the installed package to make cross-compilation easier for projects that use mdspan as dependency. 
Currently if you `make install` mdspan on a 64-bit machine, you will not be able to `find_package` it on a 32-bit system even though the installed files are just header files copied from the source directory.

See https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#command:write_basic_package_version_file for more details on the `ARCH_INDEPENDENT` flag.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,6 +148,7 @@ configure_package_config_file(cmake/mdspanConfig.cmake.in
 )
 write_basic_package_version_file(${CMAKE_CURRENT_BINARY_DIR}/mdspanConfigVersion.cmake
   COMPATIBILITY SameMajorVersion
+  ARCH_INDEPENDENT
 )
 
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/mdspanConfig.cmake ${CMAKE_CURRENT_BINARY_DIR}/mdspanConfigVersion.cmake
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.13)
+cmake_minimum_required(VERSION 3.14)
 
 project(MDSpan
   VERSION 0.2.0


CMakeLists.txt
```diff
@@ -148,6 +148,7 @@ configure_package_config_file(cmake/mdspanConfig.cmake.in
 )
 write_basic_package_version_file(${CMAKE_CURRENT_BINARY_DIR}/mdspanConfigVersion.cmake
   COMPATIBILITY SameMajorVersion
+  ARCH_INDEPENDENT
```
dalg24, line 151: https://github.com/kokkos/mdspan/blob/d861a3c6eb8cf5da2baa877d7eac114e7fd42453/CMakeLists.txt#L1
`ARCH_DEPENDENT` is a CMake 3.14 thing

rafal-c: Right, sorry I missed that. Can I bump the minimum required version to 3.14 in this PR or would you rather keep supporting 3.13 in which case this PR can be declined?

dalg24: I have no objection to require a more recent version but I see that it might involve some update in the CI setup
https://github.com/kokkos/mdspan/commit/183cab273ceb356091dee60b5c3f936e4850b7e5

dalg24: I am not super familiar with GH workflows but I am surprised the "latest" Ubuntu release wouldn't come with a recent enough CMake
https://github.com/kokkos/mdspan/blob/d861a3c6eb8cf5da2baa877d7eac114e7fd42453/.github/workflows/cmake.yml#L10
It might be worth trying see if it is still an issue

mhoemmen: @dalg24 Ubuntu 20.04.4 LTS (not the latest release) comes with CMake 3.16.3.

rafal-c: Based on the comment above I went ahead and updated the minimum version to 3.14. It seems that all checks have passed.

dalg24 (approved):
This is fine by me

PR #138, opened 2022-06-17
Bring extents in line with R17

crtrott:
The biggest changes are removal of defaulted constructors/assignment
and adding from span constructor.

### DIFF
--- a/include/experimental/__p0009_bits/extents.hpp
+++ b/include/experimental/__p0009_bits/extents.hpp
@@ -371,7 +371,7 @@ public:
   MDSPAN_CONDITIONAL_EXPLICIT(N != rank_dynamic())
   MDSPAN_INLINE_FUNCTION
   constexpr
-  extents(std::span<SizeType, N> const& exts) noexcept
+  extents(std::span<SizeType, N> exts) noexcept
 #if defined(_MDSPAN_USE_ATTRIBUTE_NO_UNIQUE_ADDRESS)
     : __storage_{
 #else
--- a/include/experimental/__p0009_bits/extents.hpp
+++ b/include/experimental/__p0009_bits/extents.hpp
@@ -344,8 +344,8 @@ public:
      */
   }
 
-#if MDSPAN_HAS_CXX_20
-  // TODO: check whether this works with newest NVCC, doesn't with 11.4
+#ifdef __cpp_lib_span
+  // TODO: check whether the below works with newest NVCC, doesn't with 11.4
 #ifdef __NVCC__
   // NVCC seems to pick up rank_dynamic from the wrong extents type???
   // NVCC chokes on the fold thingy here so wrote the workaround
--- a/include/experimental/__p0009_bits/extents.hpp
+++ b/include/experimental/__p0009_bits/extents.hpp
@@ -124,6 +124,7 @@ public:
   using __base_t = detail::__no_unique_address_emulation<__storage_t>;
 #endif
 
+// private members dealing with the way we internally store dynamic extents
  private:
 
   MDSPAN_FORCE_INLINE_FUNCTION _MDSPAN_CONSTEXPR_14
@@ -188,7 +189,13 @@ public:
   { }
 #endif
 
+
+// public interface:
 public:
+  /* Defined above for use in the private code
+  using rank_type = size_t;
+  using size_type = ThisSizeType;
+  */
 
   MDSPAN_INLINE_FUNCTION
   static constexpr rank_type rank() noexcept { return sizeof...(Extents); }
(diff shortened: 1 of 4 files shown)


include/experimental/__p0009_bits/extents.hpp
```diff
@@ -306,13 +329,72 @@ class extents
...
+  {
+    /* TODO: precondition check
+     * If N != rank_dynamic() is true, exts[r] equals Er for each r for which Er is a static extent, and
+     * either
+     *   - N is zero, or
+     *   - exts[r] is nonnegative and is a representable value of type size_type for all rank index r
+     */
+  }
+
+#if MDSPAN_HAS_CXX_20
```
mhoemmen: Is the point of testing `MDSPAN_HAS_CXX_20` just to work around an NVCC bug?  If so, then would you consider just replacing this with `#ifdef __NVCC__` for now, until we can do better testing for specifically supported NVCC versions?  If not, then would you consider using specifically relevant feature test macros, like `__cpp_concepts`?

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -306,13 +329,72 @@ class extents
...
+          _MDSPAN_TRAIT(is_convertible, SizeType, size_type) &&
+          _MDSPAN_TRAIT(is_nothrow_constructible, size_type, SizeType) &&
+          (N == rank() || N == rank_dynamic())
+    )
+  )
+#endif
+  MDSPAN_CONDITIONAL_EXPLICIT(N != rank_dynamic())
+  MDSPAN_INLINE_FUNCTION
+  constexpr
+  extents(std::span<SizeType, N> const& exts) noexcept
```
mhoemmen:
```suggestion
  extents(std::span<SizeType, N> exts) noexcept
```
The current version of the proposal in the main branch passes `span` by value, not const reference.

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -337,12 +419,16 @@ class extents
   MDSPAN_INLINE_FUNCTION
   static constexpr
   size_type static_extent(size_t n) noexcept {
+    // Can't do assert here since that breaks true constexpr ness
+    // assert(n<rank());
```
mhoemmen, line 423: Should we consider using `is_constant_evaluated()` here?
```c++
if (! std::is_constant_evaluated()) {
  assert(n < rank());
}
```
See e.g., https://godbolt.org/z/hYKT5c9qM
```c++
#include <cassert>
#include <iostream>
#include <random>
#include <type_traits>

int random_number()
{
    // Seed with a real random value, if available
    std::random_device r;
 
    // Choose a random mean between 1 and 6
    std::default_random_engine e1(r());
    std::uniform_int_distribution<int> uniform_dist(1, 6);
    int mean = uniform_dist(e1);
    return mean;
}

constexpr int foo(int k) {
    if (std::is_constant_evaluated()) {
        return -1;
    } else {
        assert(k + 1 > 0);
        return k - 2;
    }
}

template<int k = foo(50)>
struct Thing {
  Thing() { std::cout << "Thing: k = " << k << std::endl; }
};

int main() {
  const int i = random_number();
  std::cout << "foo(" << i << ") = " << foo(i) << std::endl;
  std::cout << "foo(" << 42 << ") = " << foo(42) << std::endl;
  Thing<> thing_one;
  return 0;
}
```

include/experimental/__p0009_bits/standard_layout_static_array.hpp
```diff
@@ -53,6 +53,9 @@
 #endif
 
 #include <array>
+#if MDSPAN_HAS_CXX_20
```
mhoemmen: Would you consider using the feature test macro `__cpp_lib_span` instead?

tests/test_extents.cpp
```diff
@@ -362,6 +362,18 @@ TEST(TestExtentsCtorStdArrayConvertibleToSizeT, test_extents_ctor_std_array_conv
   ASSERT_EQ(e.extent(1), 2);
 }
 
+#if MDSPAN_HAS_CXX_20
```
mhoemmen:
```suggestion
#ifdef __cpp_lib_span
```

include/experimental/__p0009_bits/standard_layout_static_array.hpp
```diff
@@ -510,6 +590,22 @@ struct __standard_layout_psa<_Tag, _T, _static_t, integer_sequence<_static_t>, _
       __construct_psa_from_dynamic_exts_array_tag_t<_IDynamic>,
       array<_U, _NDynamic> const &) noexcept {}
 
+#if MDSPAN_HAS_CXX_20
```
mhoemmen:
```suggestion
#ifdef __cpp_lib_span
```

include/experimental/__p0009_bits/standard_layout_static_array.hpp
```diff
@@ -404,6 +459,31 @@ struct __standard_layout_psa<
             __next_t(__construct_psa_from_dynamic_exts_array_tag_t<_IDynamic + 1>{},
                      __vals)) {}
 
+#if MDSPAN_HAS_CXX_20
```
mhoemmen:
```suggestion
#ifdef __cpp_lib_span
```

include/experimental/__p0009_bits/standard_layout_static_array.hpp
```diff
@@ -254,6 +257,58 @@ struct __standard_layout_psa<
 #endif
   { }
 
+#if MDSPAN_HAS_CXX_20
```
mhoemmen:
```suggestion
#ifdef __cpp_lib_span
```

mhoemmen (dismissed):
Approving so you can merge right away, but there is one point of divergence from the proposal: the `extents` constructor that takes a `span` should take it by value, not const reference.

include/experimental/__p0009_bits/extents.hpp
crtrott: span is C++20 ... so the constructor doesn't work if we don't have that :-)

include/experimental/__p0009_bits/extents.hpp
crtrott: yeah will change

include/experimental/__p0009_bits/extents.hpp
crtrott: yeah we should try that. Just need to come up with a way in the constructors to actually spell the whole condition. But we can do that in another PR later.

mhoemmen (approved):

PR #140, opened 2022-06-21
R17 updates: layout_stride

crtrott:


tests/test_layout_stride.cpp
```diff
@@ -107,7 +107,8 @@ TEST(TestLayoutStrideListInitialization, test_list_initialization) {
   ASSERT_EQ(m.extents().extent(1), 32);
   ASSERT_EQ(m.stride(0), 1);
   ASSERT_EQ(m.stride(1), 128);
-  ASSERT_EQ(m.strides(), (std::array<std::size_t, 2>{1, 128}));
+  ASSERT_EQ(m.strides()[0], 1);
+  ASSERT_EQ(m.strides()[1], 128);
```
dalg24, line 111: Why did you change this?  Can we not compare collections?

include/experimental/__p0009_bits/submdspan.hpp
```diff
@@ -397,7 +397,8 @@ constexpr auto _submdspan_impl(
   auto map = _handled.make_layout_mapping(src.mapping());
   auto acc_pol = typename AP::offset_policy(src.accessor());
   return mdspan<
-    ET, decltype(map.extents()), typename decltype(_handled)::layout_type, decltype(acc_pol)
+    ET, remove_const_t<remove_reference_t<decltype(map.extents())>>,
+        typename decltype(_handled)::layout_type, remove_const_t<remove_reference_t<decltype(acc_pol)>>
```
dalg24, line 401: Why is it not `remove_cv_t`?

tests/test_layout_stride.cpp
crtrott: nope it now returns span

include/experimental/__p0009_bits/submdspan.hpp
crtrott: we don't have volatile on this thing or?

include/experimental/__p0009_bits/mdspan.hpp
```diff
@@ -111,40 +115,78 @@ class mdspan
...
+  MDSPAN_FUNCTION_REQUIRES(
+    (MDSPAN_INLINE_FUNCTION constexpr),
+    mdspan, , ,
+    /* requires */ (
+       (rank_dynamic() > 0) &&
+       _MDSPAN_TRAIT(is_default_constructible, pointer) &&
+       _MDSPAN_TRAIT(is_default_constructible, mapping_type) &&
+       _MDSPAN_TRAIT(is_default_constructible, accessor_type)
+     )
+   )
```
mhoemmen: Does `= default` not work here?  I was wondering why you needed `MDSPAN_HAS_CXX_20`.

include/experimental/__p0009_bits/mdspan.hpp
```diff
@@ -111,40 +115,78 @@ class mdspan
...
-      _MDSPAN_TRAIT(is_constructible, extents_type, SizeTypes...) &&
+      _MDSPAN_FOLD_AND(_MDSPAN_TRAIT(is_nothrow_constructible, size_type, SizeTypes) /* && ... */) &&
+      ((sizeof...(SizeTypes) == rank()) || (sizeof...(SizeTypes) == rank_dynamic())) &&
       _MDSPAN_TRAIT(is_constructible, mapping_type, extents_type) &&
       _MDSPAN_TRAIT(is_default_constructible, accessor_type)
     )
   )
   MDSPAN_INLINE_FUNCTION
   explicit constexpr mdspan(pointer p, SizeTypes... dynamic_extents)
     // TODO @proposal-bug shouldn't I be allowed to do `move(p)` here?
```
mhoemmen, line 144: You should be able to do `move(p)` here, as the accessor policy requirements say that `is_nothrow_move_constructible_v<A::pointer>` is `true`.

mhoemmen, line 144:
```suggestion
```
You should be able to do `move(p)` here, as the accessor policy requirements say that `is_nothrow_move_constructible_v<pointer>` is `true`.

include/experimental/__p0009_bits/mdspan.hpp
```diff
@@ -319,31 +395,26 @@ mdspan(ElementType*, SizeTypes...)
 
 MDSPAN_TEMPLATE_REQUIRES(
   class Pointer,
-  (!_MDSPAN_TRAIT(is_array, Pointer))
+  (_MDSPAN_TRAIT(is_pointer, std::remove_reference_t<Pointer>))
 )
-mdspan(const Pointer&) -> mdspan<std::remove_pointer_t<Pointer>, extents<size_t>>;
+mdspan(Pointer&&) -> mdspan<std::remove_pointer_t<std::remove_reference_t<Pointer>>, extents<size_t>>;
```
mhoemmen, line 392: Oh wait, we wanted `const Pointer&` still, right?

include/experimental/__p0009_bits/mdspan.hpp
crtrott: How do we attach sfinae requirements to implement the constraints on a non-tempalted default constructor without C++20 requires?

include/experimental/__p0009_bits/mdspan.hpp
crtrott: yeah it doesn't matter really, Pointer&& does work here too. But I think people are getting worked up over stuff which doesn't matter. Since this is ONLY the deduction guide. I.e. what actually happens is determined by the constructor. This thing is only used to figure out template parameters and then a constructor is called ...

include/experimental/__p0009_bits/mdspan.hpp
mhoemmen: Ah, yes, that sounds hard.  It's probably best not to worry about it in the back-port.  Otherwise you would have to specialize the whole class just for the default constructor.

include/experimental/__p0009_bits/mdspan.hpp
mhoemmen: Yup, I agree -- I saw the e-mail list discussion.

dalg24 (approved):

include/experimental/__p0009_bits/mdspan.hpp
```diff
@@ -111,76 +115,113 @@ class mdspan
   //--------------------------------------------------------------------------------
   // [mdspan.basic.cons], mdspan constructors, assignment, and destructor
 
+#if !MDSPAN_HAS_CXX_20
   MDSPAN_INLINE_FUNCTION_DEFAULTED constexpr mdspan() = default;
+#else
```
mhoemmen, line 120: Per previous discussion: Backporting the requirements to pre-C++20 would likely require duplicating and specializing the entire `mdspan` class.  That wouldn't be worth the benefit (since the class wouldn't compile anyway if the requirements weren't met).

mhoemmen (approved):

PR #141, opened 2022-06-23
Add explicit to mdspan deduction guide

mhoemmen:
`mdspan(ElementType*, SizeTypes...)` deduction guide has an `explicit` in P0009, but not in the implementation.

@crtrott @dalg24 @nliber


mhoemmen:
Christian convinced me that `explicit` would have no effect here.

PR #147, opened 2022-06-28
Rename pointer data size type contiguous

crtrott:

PR #149, opened 2022-06-28
Fix nvc++ test build error

### LINKED ISSUE #148: nvc++ build error test_contiguous_layouts.cpp
Building `tests/test_contiguous_layouts.cpp` with nvc++ 22.5-0 fails, because the type lists are too long.  Work-around is to split up the type lists.

Errors look like this (with much more text than this):
```
".../tests/googletest-src/googletest/include/gtest/internal/gtest-internal.h", line 741: error: excessive recursion at instantiation of function "testing::internal::TypeParameterizedTest<Fixture, TestSel, Types>::Register [with Fixture=TestLayoutConversion, 
```

mhoemmen:
Building tests/tests_contiguous_layouts.cpp fails with nvc++ 22.5-0.  This commit fixes that.

Fixes #148.

### DIFF
--- a/tests/test_contiguous_layouts.cpp
+++ b/tests/test_contiguous_layouts.cpp
@@ -173,10 +173,12 @@ using test_layout_conversion =
              std::integer_sequence<size_t, AllSizes...>,
              std::integral_constant<bool, Implicit>>;
 
+// We separate layout_conversion_test_types into multiple type lists
+// in order to avoid an nvc++ compiler error (recursion too deep).
 
-using layout_conversion_test_types =
+// Left <=> Left conversion
+using layout_conversion_test_types_left_to_left =
   ::testing::Types<
-    // Left <=> Left conversion
      test_layout_conversion<stdex::layout_left , stdex::layout_left , stdex::extents<size_t,dyn>,                 stdex::extents<size_t,5>,               true, 5>
     ,test_layout_conversion<stdex::layout_left,  stdex::layout_left , stdex::extents<size_t>,                    stdex::extents<size_t>,                true>
     ,test_layout_conversion<stdex::layout_left , stdex::layout_left , stdex::extents<size_t,5>,                   stdex::extents<size_t,dyn>,             false, 5>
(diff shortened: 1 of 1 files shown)


tests/test_contiguous_layouts.cpp
```diff
@@ -335,7 +357,39 @@ struct TestLayoutConversion<
   }
 };
 
-TYPED_TEST_SUITE(TestLayoutConversion, layout_conversion_test_types);
+// We can't reuse TestLayoutConversion directly for multiple suites,
+// because this results in duplicate test names.
+// However, creating an alias for each suite works around this.
+
+TYPED_TEST_SUITE(TestLayoutConversion, layout_conversion_test_types_left_to_left);
```
crtrott, line 364: Why not use an alias for this too, so that the testnames show up systematically?

tests/test_contiguous_layouts.cpp
```diff
@@ -365,7 +419,7 @@ TYPED_TEST(TestLayoutConversion, implicit_conversion) {
   #endif
     map1 = typename TestFixture::map_1_t(map2);
 
-  for(size_t r=0; r<this->exts1.rank(); r++) {
+  for(size_t r=0; r != this->exts1.rank(); ++r) {
```
crtrott, line 422: what is this?

crtrott (changes requested):

nliber (approved):

tests/test_contiguous_layouts.cpp
mhoemmen: It prevents an nvc++ build warning.

tests/test_contiguous_layouts.cpp
crtrott: Mark said nvc++ fails if he uses an alias here too.

crtrott (approved):

tests/test_contiguous_layouts.cpp
mhoemmen: Right -- not sure why.  I'll experiment a bit once PR #147 lands.  Thanks!

PR #156, opened 2022-07-10
Fixed installation on other than Ubuntu GNU/Linux distributions

xvitaly:
Fixed installation on other than Ubuntu GNU/Linux distributions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,8 @@ project(MDSpan
   LANGUAGES CXX
 )
 
+include(GNUInstallDirs)
+
 ################################################################################
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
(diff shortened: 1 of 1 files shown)


mhoemmen:
@xvitaly Thanks for the fix!  I'm not a CMake expert so I don't feel qualified to evaluate this change.  It looks harmless to test, though.  Does "other than Ubuntu GNU/Linux distributions" mean other Linux distributions (on which ones did you test?) or other operating systems entirely?

xvitaly:
>  Does "other than Ubuntu GNU/Linux distributions" mean other Linux distributions (on which ones did you test?)

Fedora Linux for example. Ubuntu uses `/usr/lib`, Fedora and most other Linux distributions - `/usr/lib64`. GNUInstallDirs will automatically resolve such issues.

> or other operating systems entirely?

Now it will work on other operating systems too, including BSD.

mhoemmen (approved):
Thanks for the fix! : - D  The CMake `install` documentation here:

https://cmake.org/cmake/help/latest/command/install.html?highlight=cmake_install_includedir

suggests that this PR would not change the CMake version requirements.

PR #159, opened 2022-07-15
submdspan: static slices

mhoemmen:
1. When `submdspan` is given a `tuple<integral_constant<T0, Value0>, integral_constant<T1, Value1>>` as a slice specifier, preserve the compile-time values in the returned `mdspan`'s `extents_type`.
2. When `submdspan` is given an `integral_constant<T, Val>` as a slice specifier, preserve the compile-time values in the returned `mdspan`'s `extents_type`.
 
Change (1) can turn a dynamic extent into a static extent.  For example, for a rank-3 `dextents<int, 3>` mdspan `m`, `submdspan(m, full_extent, full_extent, tuple{integral_constant<int, 1>{}, integral_constant<int, 3>{}})` has `extents<int, dynamic_extent, dynamic_extent, 2>`.  Since the mapping is templated on the `extents` type, a custom layout could bake all the extents information into the type as non-type template parameters, rather than relying on `constexpr` for compile-time evaluation.  This could be helpful when crossing a compilation unit boundary without link-time optimization, for example.

Change (2) only affects the offset computation.  It might still be helpful for specific custom layouts, though.  It's also an easy change that just imitates (1).

### DIFF
--- a/include/experimental/__p0009_bits/submdspan.hpp
+++ b/include/experimental/__p0009_bits/submdspan.hpp
@@ -75,6 +75,20 @@ MDSPAN_INLINE_FUNCTION constexpr
 __slice_wrap<OldExtent, OldStaticStride, size_t>
 __wrap_slice(size_t val, size_t ext, size_t stride) { return { val, ext, stride }; }
 
+template <size_t OldExtent, size_t OldStaticStride, class IntegerType, IntegerType Value0>
+MDSPAN_INLINE_FUNCTION constexpr
+__slice_wrap<OldExtent, OldStaticStride, std::integral_constant<IntegerType, Value0>>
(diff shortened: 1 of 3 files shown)


crtrott (changes requested):
Can you add a test for 3D with the static slice happening in the middle, feels like first and last might be special, otherwise looks good.

mhoemmen:
@crtrott wrote:
> Can you add a test for 3D with the static slice happening in the middle, feels like first and last might be special, otherwise looks good.

Thanks Christian! : - )  I added more 3-D tests, including

* (tuple, full, tuple),
* (full, tuple, full),
* (full, full, index), and
* (full, index, full).

I also increased the variety of layouts exercised, and added 1-D tests.

crtrott (approved):

nliber (approved):

PR #162, opened 2022-07-28
Add swap and empty

crtrott:


tests/test_mdspan_swap.cpp
```diff
@@ -0,0 +1,198 @@
...
+    __MDSPAN_DEVICE_ASSERT_EQ(m2.data_handle().data, data2);
+    __MDSPAN_DEVICE_ASSERT_EQ(m2.rank(), 2);
+    __MDSPAN_DEVICE_ASSERT_EQ(m2.rank_dynamic(), 2);
+    __MDSPAN_DEVICE_ASSERT_EQ(m2.extent(0), 4);
+    __MDSPAN_DEVICE_ASSERT_EQ(m2.extent(1), 3);
+    __MDSPAN_DEVICE_ASSERT_EQ(m2.static_extent(0), dyn);
+    __MDSPAN_DEVICE_ASSERT_EQ(m2.static_extent(1), dyn);
+    __MDSPAN_DEVICE_ASSERT_EQ(m2.stride(0), 3);
+    __MDSPAN_DEVICE_ASSERT_EQ(m2.stride(1), 1);
+    __MDSPAN_DEVICE_ASSERT_EQ(m2.accessor().flag[0], 7);
```
dalg24: Any good reason to not use the various `operator==` instead?
Do comparisons on the extent, mapping, and accessor instead of all that block that no-one wants to proofread.

crtrott: i guess, technically this is testing mdspan interfaces not whether m2.mapping returns the right thing.

dalg24 (dismissed):

include/experimental/__p0009_bits/mdspan.hpp
```diff
@@ -339,6 +339,16 @@ class mdspan
     return static_cast<size_type>(__impl::__size(*this));
   };
 
+  MDSPAN_INLINE_FUNCTION constexpr bool empty() const noexcept {
+    return size() == 0;
```
mhoemmen: I think `empty()` has different preconditions than `size() == 0` -- it should be possible to call `empty()` even if the `size()` can't be computed, no?

tests/foo_customizations.hpp
```diff
@@ -0,0 +1,225 @@
...
+  public:
+    using extents_type = Extents;
+    using index_type = typename extents_type::index_type;
+    using size_type = typename extents_type::size_type;
+    using rank_type = typename extents_type::rank_type;
+    using layout_type = layout_foo;
+  private:
+
+    static_assert(std::experimental::detail::__is_extents_v<extents_type>,
+                  "std::experimental::layout_right::mapping must be instantiated with a specialization of std::experimental::extents.");
```
mhoemmen:
```suggestion
                  "layout_foo::mapping must be instantiated with a specialization of std::experimental::extents.");
```

tests/test_mdspan_swap.cpp
```diff
@@ -0,0 +1,123 @@
...
+namespace stdex = std::experimental;
+_MDSPAN_INLINE_VARIABLE constexpr auto dyn = stdex::dynamic_extent;
+
+
+void test_mdspan_std_swap_static_extents() {
+  size_t* errors = allocate_array<size_t>(1);
+  errors[0] = 0;
+
+  dispatch([=] _MDSPAN_HOST_DEVICE () {
+    int data1[12] = {1,2,3,4,5,6,7,8,9,10,11,12};
```
mhoemmen, line 18: Would you consider making the data differ, and testing that the swap was shallow?

mhoemmen (changes requested):
Thanks!  Just a couple typos.

include/experimental/__p0009_bits/mdspan.hpp
mhoemmen:
```c++
MDSPAN_INLINE_FUNCTION constexpr bool empty() const noexcept {
  for(std::size_t r = 0; r < rank(); ++r) {
    if(extent(r) == 0) {
      return true;
    }
  }
  return false;
}
```

nliber (dismissed):

include/experimental/__p0009_bits/mdspan.hpp
crtrott: we defined it explicitly as Effects: Equivalent to return size() == 0

crtrott: ah wait that was changed in R1, uff this is gonna be poopy

tests/test_mdspan_swap.cpp
crtrott: sure

crtrott: actually not sure why that would be different from the pointer comparison

crtrott: done

tests/foo_customizations.hpp
crtrott: done

include/experimental/__p0009_bits/mdspan.hpp
crtrott: done

mhoemmen (approved):

nliber (approved):

PR #168, opened 2022-07-29
Fix mdspan::size() return type & precondition

### LINKED ISSUE #167: mdspan::size() should return size_t, not size_type
https://github.com/kokkos/mdspan/blob/d56cb92d27a78be0e64efc7788697af1f91ae695/include/experimental/__p0009_bits/mdspan.hpp#L342

Per P0009R18, the version WG21 voted into C++ this Monday, `mdspan::size()` should return `size_t`, not `size_type`.  This is probably not a spec typo, because the function has the following precondition:

> Precondition: The size of the multidimensional index space `extents()` is representable as a value of type `size_t` (**[basic.fundamental]**).

The function needs two changes.

1. Compute the product of the extents using `size_t` (regardless of `size_type` or `index_type`)
2. Return `size_t`, not `size_type`

This was reported by the RAPIDS RAFT team (thanks!).

(Please also check `mdarray::size()` -- thanks!)

mhoemmen:
1. Make `mdspan::size()` return `size_t` (per P0009R18), instead of `size_type`.

2. Make sure that `mdspan::size()` does not overflow unless the product of the extents is not representable as a value of type `size_t`.

3. Add tests for both (1) and (2).

Fixes #167.


crtrott (approved):

mhoemmen:
Thanks Christian! : - D

Issue #169, opened 2022-08-01
Questions on using accessors for restricted accesses


roastduck:
Hi, in Section 2.7 of the proposal, I see one of the design goal of custom accessors is to apply `restrict` (non-aliasing hint) to memory accesses (if a C++ implementation supports it). But I have not figured out how to implement such an accessor. Can someone provide an example?

The problem is that I cannot simply return a `restrict`ed pointer *for each item* in a `mdspan`, because the semantics I want is a `mdspan` `a` is never an alias of another `mdspan` or ordinary pointer `b`, but items inside `a` can be aliases, for example `a[3]` and `a[i]` can be aliases.

mhoemmen:
Greetings and thanks for your interest!  Your use case is different than what we envisioned for a "no-aliasing" accessor.  May we ask what data types you have in mind for the elements of the array `a`?

I'm not a C expert, but I'm not 100% convinced that your use case is not undefined behavior.  I think it might be legal use of C `restrict`, because `a+3` and `a+i` are both "based on" `a` (in the language of the C Standard), but it's definitely not what we had in mind.

There are only two places one could apply `restrict` (or `__restrict__`, `__restrict`, etc.) to a custom accessor:

1. to the accessor's `data_handle_type`, or
2. to the accessor's `reference`.

Regarding (1), the resulting custom accessor would look almost like `default_accessor`, except that `data_handle_type` would be a `restrict` pointer.  The offset policy would actually be `default_accessor` (so that offset access is not marked `restrict`, since it's aliasing access).  Assuming that what you want to accomplish is compatible with `restrict`, this is probably what you would want to do.

Regarding (2), some compilers, [such as GCC](https://gcc.gnu.org/onlinedocs/gcc/Restricted-Pointers.html), permit applying this keyword to references.  However, I don't think this would accomplish your goal.  If the accessor's `reference` type is marked `restrict`, then the reference can't alias another reference in the local context.  The accessor has no way of knowing whether it will be used in code that refers to both `a[3]` and `a[i]`.  If _you_ know that you won't be doing that, then you can always apply `restrict` to the reference by hand in the code.

In summary: write a custom accessor that looks almost like `default_accessor`, except that `data_handle_type` is a `restrict` pointer.  Make sure that the offset policy does _not_ use `restrict`.

roastduck:
[...]

PR #172, opened 2022-08-04
Add restrict_accessor example

mhoemmen:
This example shows how to write a custom mdspan accessor that uses the `restrict` (or `__restrict` or `__restrict__`) attribute to optimize unaliased array access.  It comes from my comments on Issue https://github.com/kokkos/mdspan/issues/169 .


examples/restrict_accessor/restrict_accessor.cpp
```diff
@@ -0,0 +1,298 @@
...
+    class OtherElementType,
+    /* requires */ (std::is_convertible<OtherElementType(*)[], element_type(*)[]>::value)
+    )
+  constexpr restrict_accessor(restrict_accessor<OtherElementType>) noexcept {}
+
+  constexpr reference access(data_handle_type p, size_t i) const noexcept {
+    return p[i];
+  }
+  constexpr typename offset_policy::data_handle_type
+  offset(data_handle_type p, size_t i) const noexcept {
```
crtrott, line 114: Actually an interesting question: should we throw/abort in here? Because getting an alias to a "I will not be aliased during my lifetime" thing seems iffy?

crtrott (commented):
Generally like this.

examples/restrict_accessor/restrict_accessor.cpp
mhoemmen: @crtrott An excellent question!  My view about offset is that it should behave as the language behaves.  It's legal in C (and extensions of C with the `restrict` keyword or attribute) to assign from a `restrict` pointer to a non-`restrict` pointer.  It's meaningful and necessary, because of code that accesses the elements of a `restrict` pointer-to-array using a pointer.  For example:
```c++
void assign_to_elements_of_x(int* restrict x, int n) {
  int* beg = x;
  int* end = x+n;
  while(beg != end) {
    *beg++ = 42;
  }
}
```

mhoemmen: Another example:
```c++
void compute_x(int* restrict x, int n) {
  span<int> x_whole{x, n};
  base_case(x_whole);
  span<int> x_left{x, n/2};
  span<int> x_right{x + n/2, n/2};
  compute_left(x_left);
  compute_right(x_right);
}
```

examples/restrict_accessor/restrict_accessor.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+template<class ElementType>
+void warmup(const index_type n, const ElementType* x, const ElementType* y, ElementType* z)
+{
+  constexpr std::size_t num_warmup_trials = 10;
+  for(std::size_t trial = 0; trial < num_warmup_trials; ++trial) {
+    for (index_type i = 0; i < n; ++i) {
+      z[i] = x[i] + y[i];
+    }
+  }
```
dalg24: Any good reason not to just the benchmarks directly to do the warmup?

mhoemmen: @dalg24 lol no ; - )

mhoemmen: @dalg24 fixed; thanks!

crtrott (approved):

PR #176, opened 2022-08-18
Add aligned_accessor example

mhoemmen:
1. `aligned_accessor` uses compiler identification macros to pick from various, possibly nonstandard ways to mark a pointer (its value, and/or its type) as having a specific alignment.

2. "Aligned" benchmarks use overaligned allocation and the aforementioned way to mark a pointer as aligned.

3. "Unaligned' benchmarks access memory aligned only to the element type, by adding 1 (`sizeof(ElementType)` bytes) to the array start.


mhoemmen:
Fixes for Clang, MSVC, and Intel here: https://godbolt.org/z/vhGz3zcYn

I still need to incorporate these into the PR.

mhoemmen:
@crtrott @dalg24 @nliber I updated the PR to fix Clang, MSVC, and Intel builds.  Here is the corresponding Compiler Explorer link: https://godbolt.org/z/47dEf59rs

mhoemmen:
Test with C++14, C++17, and C++20 builds: https://godbolt.org/z/raYsGTq3P

mhoemmen:
Latest update (just pushed) includes OpenMP fixes: https://godbolt.org/z/PbMWbKMj7

examples/aligned_accessor/aligned_accessor.cpp
```diff
@@ -0,0 +1,714 @@
...
+#  define _MDSPAN_ALIGN_VALUE_ATTRIBUTE( BYTE_ALIGNMENT )
+  constexpr char align_attribute_method[] = "(none)";
+#endif
+
+template<class ElementType>
+constexpr bool
+valid_byte_alignment(const std::size_t byte_alignment,
+		     const std::size_t size_of_element)
+{
+  return byte_alignment >= alignof(ElementType); // && is a power of two
```
dalg24: compare to `bit_floor(byte_alignment)` when C++20 is available?

examples/aligned_accessor/aligned_accessor.cpp
```diff
@@ -0,0 +1,714 @@
...
+
+template<class ElementType, std::size_t byte_alignment>
+struct aligned_pointer {
+  // x86-64 ICC 2021.5.0 emits warning #3186 ("expected typedef declaration") here.
+  // No other compiler (including Clang, which has a similar type attribute) has this issue.
+  using type = ElementType* _MDSPAN_ALIGN_VALUE_ATTRIBUTE( byte_alignment );
+};
+
+template<class ElementType, std::size_t byte_alignment>
+using aligned_pointer_t = typename aligned_pointer<ElementType, byte_alignment>::type;
```
dalg24: Why do you bother with the trait and don't directly define the alias?

examples/aligned_accessor/aligned_accessor.cpp
```diff
@@ -0,0 +1,714 @@
...
+  using data_handle_type = aligned_pointer_t<ElementType, byte_alignment>;
+
+  constexpr aligned_accessor() noexcept = default;
+
+  MDSPAN_TEMPLATE_REQUIRES(
+    class OtherElementType,
+    std::size_t other_byte_alignment,
+    /* requires */ (std::is_convertible<OtherElementType(*)[], element_type(*)[]>::value && other_byte_alignment == byte_alignment)
+    )
+  constexpr aligned_accessor(aligned_accessor<OtherElementType, other_byte_alignment>) noexcept {}
```
dalg24, line 183: Why a requires clause rather than a static_assert?

examples/aligned_accessor/aligned_accessor.cpp
```diff
@@ -0,0 +1,714 @@
...
+  return TOCK();
+}
+
+// Assume that x, y, and z all have the same alignment.
+template<class ElementType, std::size_t byte_alignment>
+void add_aligned_raw_1d(const index_type n,
+			aligned_pointer_t<const ElementType, byte_alignment> x,
+			aligned_pointer_t<const ElementType, byte_alignment> y,
+			aligned_pointer_t<ElementType, byte_alignment> z,
+			std::integral_constant<std::size_t, byte_alignment> /* ba */ )
```
dalg24: Why do you bother with that argument

dalg24 (commented):
A few drive by comments

examples/aligned_accessor/aligned_accessor.cpp
mhoemmen: @dalg24 Excellent suggestion; thank you!  I should be able to use the feature test macro `__cpp_lib_int_pow2` for these C++20 functions.

examples/aligned_accessor/aligned_accessor.cpp
mhoemmen: Good point -- I must have mixed up in my head some missing C++14 thing from this C++11 feature.  I'll change it.

examples/aligned_accessor/aligned_accessor.cpp
mhoemmen: @dalg24 The `element_type` part of this imitates `default_accessor`.  It felt weird to treat byte alignment differently than `element_type`.  I won't insist, though ; - ) .

The byte alignment part could, in theory, permit conversions to lesser alignment.

examples/aligned_accessor/aligned_accessor.cpp
mhoemmen: 1. Consistency with the aligned mdspan interface
2. `aligned_pointer_t<ElementType, byte_alignment` is an alias and might just be `ElementType*`, but it might be a pointer with an attribute.  I wasn't sure whether compilers could deduce `byte_alignment` from an attribute on the pointer.

I'll experiment with removing this parameter.

mhoemmen: I was able to remove this parameter; thanks!

examples/aligned_accessor/aligned_accessor.cpp
mhoemmen: Fixed; thanks!

examples/aligned_accessor/aligned_accessor.cpp
mhoemmen: Fixed; thank you!

examples/aligned_accessor/aligned_accessor.cpp
mhoemmen: I ended up bringing back the struct, because this lets me check alignment at compile time for all uses of the alias.  (Just defining the alias directly wouldn't let me impose a `static_assert` on it for all specializations.)

examples/aligned_accessor/aligned_accessor.cpp
mhoemmen: @dalg24 Please let me know if you still want this changed; thanks!

examples/aligned_accessor/aligned_accessor.cpp
```diff
@@ -0,0 +1,737 @@
...
+
+// mfh 2022/08/08: This is based on my comment on RAPIDS RAFT issue 725:
+// https://github.com/rapidsai/raft/pull/725#discussion_r937991701
+
+namespace {
+
+using test_value_type = float;
+constexpr int min_overalignment_factor = 8;
+constexpr int min_byte_alignment =
+  static_cast<int>(min_overalignment_factor * sizeof(float));
```
dalg24: cast is pointless

examples/aligned_accessor/aligned_accessor.cpp
```diff
@@ -0,0 +1,737 @@
...
+template<class ElementType>
+using mdspan_1d =
+  stdex::mdspan<ElementType,
+		stdex::dextents<index_type, 1>,
+		stdex::layout_right,
+		stdex::default_accessor<ElementType>>;
+
+#define TICK() const auto tick = std::chrono::steady_clock::now()
+
+using double_seconds = std::chrono::duration<double, std::ratio<1>>;
```
dalg24, line 353: Why would you define an alias if you ever only use it in the TOCK macro definition?

examples/aligned_accessor/aligned_accessor.cpp
```diff
@@ -0,0 +1,737 @@
...
+					  x.data_aligned(), y.data_aligned(), z.data_aligned(),
+					  byte_alignment);
+  auto omp_aligned_simd_aligned_raw_result =
+    benchmark_add_omp_aligned_simd_aligned_raw_1d(num_trials, n,
+						  x.data_aligned(), y.data_aligned(), z.data_aligned(),
+						  byte_alignment);
+
+#endif // _OPENMP
+
+  cout << "Number of trials: " << num_trials << endl
```
dalg24, line 739: Flushing the stream excessively

examples/aligned_accessor/aligned_accessor.cpp
```diff
@@ -0,0 +1,737 @@
...
+  using std::cerr;
+  using std::endl;
+
+  if(argc != 3) {
+    cerr << "Usage: main <n> <num_trials>" << endl;
+    return -1;
+  }
+
+  int n = 10000;
+  int num_trials = 100;
```
dalg24: Why do you bother with initializing to overwrite the value below

dalg24 (commented):
Looks mostly good
Only question I have is do you care to not cast within the outer repetition loop?
Because if that is acceptable you would be able to get rid of a lot of boiler plate code

examples/aligned_accessor/aligned_accessor.cpp
mhoemmen: I'm not too worried about this, since this is the end of the benchmark and not performance-critical code.

examples/aligned_accessor/aligned_accessor.cpp
mhoemmen: It makes the definition of `TOCK` shorter and better expresses the code's intent.

mhoemmen:
@dalg24 wrote:
> Only question I have is do you care to not cast within the outer repetition loop?

The "casts" in the various `benchmark_` functions don't all have the same form.  Some of them actually create an mdspan from a pointer.  Others call `bless` on the pointer, which both casts and calls `assume_aligned`.  Performing the casts or blessings explicitly, outside the outer loop, lets each `add_*` function be as simple as possible.  This hopefully removes any unnecessary barriers to the compiler optimizing it.

This is an example, so the small amount of repetition has educational value.  Users can copy and paste as a start to adding their own benchmarks.

examples/aligned_accessor/aligned_accessor.cpp
mhoemmen: @dalg24 Fixed; thanks!

examples/aligned_accessor/aligned_accessor.cpp
mhoemmen: @dalg24 Fixed; thanks!

examples/aligned_accessor/aligned_accessor.cpp
```diff
@@ -0,0 +1,734 @@
...
+    return _MDSPAN_ASSUME_ALIGNED( ElementType, alloc.aligned_pointer, byte_alignment );
+  }
+
+  // A pointer to the first (NOT zeroth) element of the array.
+  // It is aligned only to sizeof(ElementType) bytes.
+  // All num_elements (constructor parameter) elements are valid.
+  //
+  // This exists so that we can compare performance
+  // of aligned and unaligned access.
+  ElementType* data_unaligned() const noexcept
```
crtrott: Not sure if this is what we should do here instead of returning also address to element 0. After all you are now measuring something different, like the number of cachelines/pages touched by each thread might be different. I'd rather like to see that the attribute in and of itself does something.

crtrott (changes requested):
I think we need to change the non-aligned element returning & to element 0 too, I am also wondering if we can reduce code duplication here. For example have the same benchmark function templated on a tag and the casted to type:

```c++
template<class ArrayType, class constArrayType, class ElementType,  class BenchmarkTag, class ... ConstructorArgs>
auto benchmark_add_1d(const BenchMarkTag& tag, const std::size_t num_trials,
				      const index_type n,
				      const ElementType x[],
				      const ElementType y[],
				      ElementType z[],
                                      ConstructorArgs ... cargs)
{
  TICK();
  constArrayType x2{x, cargs};
  constArrayType y2{x, cargs};
  ArrayType z2{x, cargs};
  for (std::size_t trial = 0; trial < num_trials; ++trial) {
    add(tag, x2, y2, z2);
  }
  return TOCK();
}
```

examples/aligned_accessor/aligned_accessor.cpp
mhoemmen: @crtrott My concern there was that the compiler has complete code visibility, so it could deduce that the allocation is aligned.  For the benchmark, I wanted to be able to give it an allocation that I _know_ is aligned to no more than `sizeof(ElementType)`.

mhoemmen: @crtrott I like the tag idea for the benchmarks, though!  I was a bit worried about possibly needing to do type conversion of x, y, and z on every invocation of the inner loop, but I think I could restructure the code to avoid that.

mhoemmen: @crtrott I tried out the tag idea, but I don't think it saves much code.  Each benchmark invocation still needs to name the const and nonconst array types and the tag types explicitly.

Perhaps y'all are missing the point a bit -- this is an _example_.  It's code for users to read, that teaches them how to write a custom accessor.  It's something more like a tutorial.  Reducing code duplication isn't really the point here; the point is simple, readable code that users could easily copy and paste.

mhoemmen:
@crtrott wrote:
> I think we need to change the non-aligned element returning & to element 0 too....

I have addressed this in my latest revision.

crtrott (approved):

PR #182, opened 2022-08-30
Work around __make_dynamic_extent compiler bug

### LINKED ISSUE #177: Consider __make_dynamic_extent work-around for extents deduction guide
@youyu3 encountered some issues compiling `extents` with nvcc + GCC with certain compiler versions and settings.  The issue appears to be with `__make_dynamic_extent()`.  Changing the `extents(IndexTypes...)` deduction guide so it looks like the following (not including constraints for clarity) makes it work.

```c++
template <class... IndexTypes>
extents(IndexTypes...) -> extents<size_t, size_t((IndexTypes(), -1))...>;
```

Should we consider doing this in the reference implementation as well, or at least doing this (if it works with nvcc)?

```c++
template <class... IndexTypes>
extents(IndexTypes...) -> extents<size_t, size_t((IndexTypes(), dynamic_extent))...>;
```

mhoemmen:
Some compilers can't do CTAD correctly with extents.  Replacing `__make_dynamic_extent<IndexTypes>()` with an explicit expression in the deduction guide's parameter pack expansion works around the issue.

The resulting change means that nothing in mdspan uses `__make_dynamic_extent`.  I've left `__make_dynamic_extent` defined, just in case downstream projects rely on it.

Fixes #177.  @youyu3

### DIFF
--- a/include/experimental/__p0009_bits/dynamic_extent.hpp
+++ b/include/experimental/__p0009_bits/dynamic_extent.hpp
@@ -53,20 +53,6 @@ namespace experimental {
 
 _MDSPAN_INLINE_VARIABLE constexpr auto dynamic_extent = std::numeric_limits<size_t>::max();
 
-namespace detail {
-
-template <class>
-constexpr auto __make_dynamic_extent() {
-  return dynamic_extent;
-}
-
-template <size_t>
-constexpr auto __make_dynamic_extent_integral() {
-  return dynamic_extent;
-}
-
-} // end namespace detail
-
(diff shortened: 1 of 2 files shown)


crtrott:
Lets delete that thing. Nobody should start to rely on things in detail with two underscores as start characters. Better they learn that now.

mhoemmen:
@crtrott I removed `__make_dynamic_extent` and `__make_dynamic_extent_integral` -- thanks!

crtrott (approved):

PR #184, opened 2022-09-02
Fixes issue #175

### LINKED ISSUE #175: layout_{left,right}::mapping(const extents_type&) needs MDSPAN_INLINE_FUNCTION
The `mapping(const extents_type&)` constructor in both `layout_left` and `layout_right` needs blessing with `MDSPAN_INLINE_FUNCTION`.  @youyu3 discovered this when attempting to construct an mdspan on device, with g++ 11.2.0 and nvcc 11.7.  It looks like the analogous `layout_stride::mapping` constructor already has this blessing.

youyu3:
* Declares a few functions in `layout_{left|right}` as `_MDSPAN_HOST_DEVICE` to avoid compile-time warning messages when built with nvcc without the `--expt-relaxed-constexpr` flag

Fixes #175 .

### DIFF
--- a/include/experimental/__p0009_bits/layout_left.hpp
+++ b/include/experimental/__p0009_bits/layout_left.hpp
@@ -72,6 +72,7 @@ class layout_left::mapping {
     struct __rank_count {};
 
     template <size_t r, size_t Rank, class I, class... Indices>
+    _MDSPAN_HOST_DEVICE
     constexpr index_type __compute_offset(
       __rank_count<r,Rank>, const I& i, Indices... idx) const {
       return __compute_offset(__rank_count<r+1,Rank>(), idx...) *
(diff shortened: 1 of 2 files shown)


mhoemmen:
@crtrott We had an offline discussion about whether to use `MDSPAN_INLINE_FUNCTION` instead of `_MDSPAN_HOST_DEVICE`.  We concluded that using `_MDSPAN_HOST_DEVICE` is fine, for the following reasons.

1. The tests use it (for the lambdas given to `dispatch`).
2. The functions in question don't need the `inline` keyword.  It's generally desirable to have a way to spell `__host__ __device__` without saying `inline` as well.

Given (2), I'll file an issue to rename `_MDSPAN_HOST_DEVICE` to `MDSPAN_HOST_DEVICE`, so that tests and examples can feel more confident to use it.  It's not necessarily good to force the implementation to mix `inline` with `__host__ __device__`, because `inline` changes ODR semantics.  For a header-only library, this doesn't matter quite so much, but users might pick up on the habit.

mhoemmen (approved):

PR #204, opened 2022-11-03
Fix build on MacOS: don't use std::aligned_alloc

hcho3:
Apple Clang does not yet offer `std::aligned_alloc`, even with C++14 enabled. (https://forum.qt.io/topic/100242/aligned_alloc-is-not-available-despite-c-17-flag)


mhoemmen (approved):
Thanks for the fix!  I don't have a way to test on a Mac straightforwardly, but I have tested that branch of the `#elif`, so this should work fine.

PR #211, opened 2022-11-14
use setup-oneapi instead of manual install

rscohn2:


crtrott:
Hi,

I am slightly hesitant to refer to rscohn2/setup-oneapi@v0 in this thing. Is there a possibility that this setup-oneapi may be provided by one of the official Intel accounts?

Cheers
Christian

rscohn2:
>  Is there a possibility that this setup-oneapi may be provided by one of the official Intel accounts?

I hope it will happen eventually, but there are not prospects of it happening soon. I did this on my own time and the intel process for open sourcing is heavyweight. I will close the PR. The code you have now works fine, it just requires manually updating the version. I used your project as a test for the action.

PR #212, opened 2022-11-22
New extents implementation

crtrott:
This replaces the extents implementation with something significanlty simpler. Need to do some more performance testing, but this is ready for review.


crtrott:
You may just want to look at the new extents file not in the diff. I did just replace the file after all.

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,466 +13,519 @@
...
+#if MDSPAN_HAS_CXX_17
+    if constexpr (r == R)
+      return FirstExt;
+    else
+      return static_array_impl<R + 1, T, Extents...>::template get<r>();
+#else
+    if (r == R)
+      return FirstExt;
+    else
+      return static_array_impl<R + 1, T, Extents...>::value(r);
```
dalg24:
```suggestion
    return value(r);
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,466 +13,519 @@
...
+template <size_t R, class T, T FirstExt, T... Extents>
+struct static_array_impl<R, T, FirstExt, Extents...> {
+  MDSPAN_INLINE_FUNCTION
+  constexpr static T value(size_t r) {
+    if (r == R)
+      return FirstExt;
+    else
+      return static_array_impl<R + 1, T, Extents...>::value(r);
+  }
+  template <size_t r> MDSPAN_INLINE_FUNCTION constexpr static T get() {
```
dalg24, line 78: I really hate that guy.  Kinda wish you had overloaded the same function to take either an integral type or std::integral_constant

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,466 +13,519 @@
...
+  MDSPAN_INLINE_FUNCTION
+  constexpr static size_t get(size_t r) {
+    if (r > R)
+      return FirstVal + index_sequence_scan_impl<R + 1, Values...>::get(r);
+    else
+      return 0;
+  }
+  template <size_t r> MDSPAN_INLINE_FUNCTION constexpr static size_t get() {
+    return r > R ? FirstVal + index_sequence_scan_impl<R + 1, Values...>::get(r)
+                 : 0;
```
dalg24: Might want the if constexpr here?

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,466 +13,519 @@
...
-  constexpr __storage_t const& __storage() const noexcept {
-#if defined(_MDSPAN_USE_ATTRIBUTE_NO_UNIQUE_ADDRESS)
-    return __storage_;
 #else
-    return this->__base_t::__ref();
+  MDSPAN_INLINE_FUNCTION
+  constexpr static size_t get(size_t r) { return R > r ? FirstVal : 0; }
 #endif
+  template <size_t r> MDSPAN_INLINE_FUNCTION constexpr static size_t get() {
+    return R > r ? FirstVal : 0;
```
dalg24: Same if constexpr comment

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,466 +13,519 @@
...
+    return 0;
   }
+};
 
-  template <class, size_t...>
-  friend class extents;
-
-  template <class OtherIndexType, size_t... OtherExtents, size_t... Idxs>
+// Need this to have fully empty class maybe_static_array with all entries
+// static Rather specialize this small helper class on size 0, then the whole
```
dalg24: add a "." to terminate your sentence.  I expect that is clang-format doing here and you had two separate lines.

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,466 +13,519 @@
...
+    for (size_t r = 0; r < N; r++)
+      m_dyn_vals[r] = static_cast<TDynamic>(vals[r]);
+  }
 
-  // Default constructor
-  MDSPAN_INLINE_FUNCTION_DEFAULTED constexpr extents() noexcept = default;
+  MDSPAN_TEMPLATE_REQUIRES(class T, size_t N,
+                           /* requires */ (N == m_size_dynamic && N == 0))
+  MDSPAN_INLINE_FUNCTION
+  constexpr maybe_static_array(const std::array<T, N> &) {}
```
dalg24: I learnt that you can do std::array<T, 0>
OK

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -484,34 +537,21 @@ using dextents = typename detail::__make_dextents<IndexType, Rank>::type;
...
-struct __is_extents : ::std::false_type {};
+template <class T> struct __is_extents : ::std::false_type {};
 
 template <class IndexType, size_t... ExtentsPack>
-struct __is_extents<::std::experimental::extents<IndexType, ExtentsPack...>> : ::std::true_type {};
+struct __is_extents<::std::experimental::extents<IndexType, ExtentsPack...>>
+    : ::std::true_type {};
 
 template <class T>
 static constexpr bool __is_extents_v = __is_extents<T>::value;
```
dalg24: Why "static" and not "inline"?

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,466 +13,519 @@
...
+    return !(lhs == rhs);
+  }
+#endif
 };
 
 namespace detail {
 
-template <class IndexType, size_t Rank, class Extents = ::std::experimental::extents<IndexType>>
+template <class IndexType, size_t Rank,
+          class Extents = ::std::experimental::extents<IndexType>>
```
dalg24, line 543: What's the point of the default argument?

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,466 +13,519 @@
...
+            std::integral_constant<size_t, 0>(), other)) {}
+
+  template <class OtherIndexType, size_t... OtherExtents>
+  MDSPAN_INLINE_FUNCTION friend constexpr bool
+  operator==(const extents &lhs,
+             const extents<OtherIndexType, OtherExtents...> &rhs) noexcept {
+    bool value = true;
+    for (size_type r = 0; r < m_rank; r++)
+      value &= rhs.extent(r) == lhs.extent(r);
+    return value;
```
dalg24, line 526: Why not something more simple like
```suggestion
    for (size_type r = 0; r < m_rank; r++)
      if (rhs.extent(r) != lhs.extent(r))
        return false;
    return true;
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,466 +13,519 @@
...
+      class... OtherIndexTypes,
+      /* requires */ (
+          _MDSPAN_FOLD_AND(_MDSPAN_TRAIT(is_convertible, OtherIndexTypes,
+                                         index_type) /* && ... */) &&
+          _MDSPAN_FOLD_AND(_MDSPAN_TRAIT(is_nothrow_constructible, index_type,
+                                         OtherIndexTypes) /* && ... */) &&
+          (sizeof...(OtherIndexTypes) == m_rank ||
+           sizeof...(OtherIndexTypes) == m_rank_dynamic)))
+  MDSPAN_INLINE_FUNCTION
+  constexpr extents(OtherIndexTypes... dynvals)
```
dalg24:
```suggestion
  constexpr explicit extents(OtherIndexTypes... dynvals) noexcept
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,466 +13,519 @@
...
+  MDSPAN_TEMPLATE_REQUIRES(
+      class OtherIndexType, size_t N,
+      /* requires */
+      (
+          //    _MDSPAN_TRAIT(is_convertible, OtherIndexType, index_type) &&
+          //    _MDSPAN_TRAIT(is_nothrow_constructible, index_type,
+          //    OtherIndexType) &&
+          (N == m_rank || N == m_rank_dynamic)))
+  MDSPAN_INLINE_FUNCTION
+  constexpr extents(const array<OtherIndexType, N> &exts)
```
dalg24:
```suggestion
  constexpr explicit(N != rank_dynamic()) extents(const array<OtherIndexType, N> &exts) noexcept
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,466 +13,519 @@
...
+
+#ifdef __cpp_lib_span
+  MDSPAN_TEMPLATE_REQUIRES(
+      class OtherIndexType, size_t N,
+      /* requires */
+      (_MDSPAN_TRAIT(is_convertible, OtherIndexType, index_type) &&
+       _MDSPAN_TRAIT(is_nothrow_constructible, index_type, OtherIndexType) &&
+       (N == m_rank || N == m_rank_dynamic)))
+  MDSPAN_INLINE_FUNCTION
+  constexpr extents(const span<OtherIndexType, N> &exts)
```
dalg24:
```suggestion
  constexpr explicit(N != rank_dynamic()) extents(const span<OtherIndexType, N> &exts) noexcept
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,466 +13,519 @@
...
   MDSPAN_INLINE_FUNCTION
-  friend constexpr bool operator==(extents const& lhs, extents<OtherIndexType, RHS...> const& rhs) noexcept {
-    return lhs._eq_impl(
-      rhs, std::integral_constant<bool, (sizeof...(RHS) == rank())>{},
-      make_index_sequence<sizeof...(RHS)>{}
-    );
-  }
+  constexpr static rank_type rank() { return m_rank; }
+  MDSPAN_INLINE_FUNCTION
+  constexpr static rank_type rank_dynamic() { return m_rank_dynamic; }
```
dalg24:
```suggestion
  constexpr static rank_type rank_dynamic() noexcept { return m_rank_dynamic; }
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,466 +13,519 @@
...
+
+public:
   MDSPAN_INLINE_FUNCTION
-  friend constexpr bool operator==(extents const& lhs, extents<OtherIndexType, RHS...> const& rhs) noexcept {
-    return lhs._eq_impl(
-      rhs, std::integral_constant<bool, (sizeof...(RHS) == rank())>{},
-      make_index_sequence<sizeof...(RHS)>{}
-    );
-  }
+  constexpr static rank_type rank() { return m_rank; }
```
dalg24:
```suggestion
  constexpr static rank_type rank() noexcept { return m_rank; }
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,466 +13,519 @@
...
 
-#if !(MDSPAN_HAS_CXX_20)
-  template<class OtherIndexType, size_t... RHS>
   MDSPAN_INLINE_FUNCTION
-  friend constexpr bool operator!=(extents const& lhs, extents<OtherIndexType, RHS...> const& rhs) noexcept {
-    return lhs._not_eq_impl(
-      rhs, std::integral_constant<bool, (sizeof...(RHS) == rank())>{},
-      make_index_sequence<sizeof...(RHS)>{}
-    );
+  constexpr index_type extent(rank_type r) const { return m_vals.value(r); }
```
dalg24:
```suggestion
  constexpr index_type extent(rank_type r) const noexcept { return m_vals.value(r); }
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,466 +13,519 @@
...
+  MDSPAN_INLINE_FUNCTION
+  constexpr static size_t static_extent(rank_type r) {
```
dalg24:
```suggestion
  constexpr static size_t static_extent(rank_type r) noexcept {
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,466 +13,519 @@
...
+    return vals_t::static_value(r);
   }
-#endif
 
-  // End of public interface
+  MDSPAN_INLINE_FUNCTION_DEFAULTED
+  constexpr extents() = default;
```
dalg24:
```suggestion
  constexpr extents() noexcept = default;
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,466 +13,519 @@
...
   MDSPAN_INLINE_FUNCTION
-  static constexpr
-  index_type __static_extent() noexcept {
-    return __storage_t::template __get_static_n<N, Default>();
+  MDSPAN_CONDITIONAL_EXPLICIT((((Extents != dynamic_extent) &&
+                                (OtherExtents == dynamic_extent)) ||
+                               ...) ||
+                              (std::numeric_limits<index_type>::max() <
+                               std::numeric_limits<OtherIndexType>::max()))
+  constexpr extents(const extents<OtherIndexType, OtherExtents...> &other)
```
dalg24:
```suggestion
  constexpr extents(const extents<OtherIndexType, OtherExtents...> &other) noexcept
```

include/experimental/__p0009_bits/extents.hpp
crtrott: at the point where I use it in the deduction guide I don't know the type I need to produce. I believe its less confusing having a default argument here, when writing:

```c++
template <class IndexType, size_t Rank>
using dextents = typename detail::__make_dextents<IndexType, Rank, extents<IndexType>>::type;
```

I.e. this meta functions is meant to produce dextents from the inputs of index type and rank.

include/experimental/__p0009_bits/extents.hpp
crtrott: I don't feel that its simpler, its also 4 lines after all. I am also wondering if the above can be easier optimized via unrolling and the constexpr nature of the values you combine, as opposed to also have the conditional in there.

crtrott:
Some perf tests:
```c++
template <class Extents, class ... Args>
void extents_construction(benchmark::State& state, Extents, int R, Args ... args) {
  for (auto _ : state) {
    double val =0;
    for(int r=0; r<R; r++) {
       Extents ext(args...);
       val += ext.extent(0);
    }
    if(val<1.0*R*10) std::terminate();
  }
}

BENCHMARK_CAPTURE(
  extents_construction, int_int_int, stdex::dextents<int,3>(), 100000000, 100, 100, 100
);
BENCHMARK_CAPTURE(
  extents_construction, array_int_3, stdex::dextents<int,3>(), 100000000, std::array<int,3>{100, 100, 100}
);
```

In -O0 the new extents are faster. In O1 and above they are the same. 

O0:
```
# stable
extents_construction/int_int_int 3966498599 ns   3957175517 ns            1
extents_construction/array_int_3 4942861061 ns   4931248140 ns            1
# new-extents-impl
extents_construction/int_int_int 1841957323 ns   1837625666 ns            1
extents_construction/array_int_3 3745596231 ns   3736797353 ns            1
```

O3:
```
#stable
extents_construction/int_int_int  108691256 ns    108436294 ns            6
extents_construction/array_int_3  108762740 ns    108507276 ns            6
#new-extents-impl
extents_construction/int_int_int  108782075 ns    108526909 ns            6
extents_construction/array_int_3  108849447 ns    108594120 ns            6
```

crtrott:
Compile time test construction:
```c++
#include<experimental/mdspan>

constexpr size_t dyn = std::experimental::dynamic_extent;
template<class IndexT, size_t R, size_t ... Args>
struct ExtentsConstructor {
  using extents_t = std::experimental::extents<IndexT, R, Args...>;
  template<class ... DynArgs>
  static constexpr size_t value (DynArgs ... args) {
    extents_t ext(R,args...);
    return ext.extent(0) + ExtentsConstructor<IndexT, R-1, Args...>::value(args...);
  }
};

template<class IndexT, size_t ... Args>
struct ExtentsConstructor<IndexT, 0, Args...> {
  template<class ... DynArgs>
  static constexpr size_t value (DynArgs ... args) { return 0; }
};

size_t foo(int val1) {
  size_t value = 0;
  value += ExtentsConstructor<int32_t, 120, 3, 4, 5>::value(val1,4,5);
  value += ExtentsConstructor<uint32_t, 120, 3, 4, 5>::value(val1,4,5);
  value += ExtentsConstructor<int64_t, 120, 3, 4, 5>::value(val1,4,5);
  value += ExtentsConstructor<uint64_t, 120, 3, 4, 5>::value(val1,4,5);
  value += ExtentsConstructor<int32_t, 120, dyn,dyn,dyn>::value(val1,val1,val1);
  value += ExtentsConstructor<uint32_t, 120, dyn,dyn,dyn>::value(val1,val1,val1);
  value += ExtentsConstructor<int64_t, 120, dyn,dyn,dyn>::value(val1,val1,val1);
  value += ExtentsConstructor<uint64_t, 120, dyn,dyn,dyn>::value(val1,val1,val1);
  return value;
}
```

Simply compile as object file:
`time g++ -O3 -c test.cpp -I${HOME}/Kokkos/mdspan/include`

GCC 11.1 Old: 5.4s New 4.8s
Clang 15 Old: 4.0s New 3.9s

So looks like compile time for construction went down.

crtrott:
Compile time access benchmark:
```c++
#include<experimental/mdspan>

constexpr size_t dyn = std::experimental::dynamic_extent;
template<class Extents, size_t R>
struct ExtentsConstructor {
  static constexpr size_t value (const Extents& ext) {
    return ext.extent(static_cast<int32_t>(R%Extents::rank())) +
           ext.extent(static_cast<int64_t>(R%Extents::rank())) +
           ext.extent(static_cast<uint32_t>(R%Extents::rank())) +
           ext.extent(static_cast<uint64_t>(R%Extents::rank())) +
            + ExtentsConstructor<Extents, R-1>::value(ext);
  }
};

template<class Extents>
struct ExtentsConstructor<Extents, 0> {
  static constexpr size_t value (const Extents&) { return 0; }
};

size_t foo(int val1) {
  size_t value = 0;
  {
    using ext_t = std::experimental::extents<int32_t, 3, 4, 5>;
    value += ExtentsConstructor<ext_t, 120>::value(ext_t(val1,4,5));
  }
  {
    using ext_t = std::experimental::extents<int64_t, 3, 4, 5>;
    value += ExtentsConstructor<ext_t, 120>::value(ext_t(val1,4,5));
  }
  {
    using ext_t = std::experimental::extents<uint32_t, 3, 4, 5>;
    value += ExtentsConstructor<ext_t, 120>::value(ext_t(val1,4,5));
  }
  {
    using ext_t = std::experimental::extents<uint64_t, 3, 4, 5>;
    value += ExtentsConstructor<ext_t, 120>::value(ext_t(val1,4,5));
  }
  {
    using ext_t = std::experimental::extents<int32_t, dyn, dyn, dyn>;
    value += ExtentsConstructor<ext_t, 120>::value(ext_t(val1,4,5));
  }
  {
    using ext_t = std::experimental::extents<int64_t, dyn, dyn, dyn>;
    value += ExtentsConstructor<ext_t, 120>::value(ext_t(val1,4,5));
  }
  {
    using ext_t = std::experimental::extents<uint32_t, dyn, dyn, dyn>;
    value += ExtentsConstructor<ext_t, 120>::value(ext_t(val1,4,5));
  }
  {
    using ext_t = std::experimental::extents<uint64_t, dyn, dyn, dyn>;
    value += ExtentsConstructor<ext_t, 120>::value(ext_t(val1,4,5));
  }
  return value;
}
```
GCC 11.1 Old: 0.95 New 0.89
Clang 15 Old: 0.72 New 0.66

crtrott:
I also ran the existing benchmarks and didn't see anything bad. So in general I think this is good to go.

include/experimental/__p0009_bits/extents.hpp
crtrott: Done

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,505 +13,602 @@
...
-#include "static_array.hpp"
-#include "standard_layout_static_array.hpp"
-#include "trait_backports.hpp" // integer_sequence, etc.
+#pragma once
+#include "dynamic_extent.hpp"
 
-#if !defined(_MDSPAN_USE_ATTRIBUTE_NO_UNIQUE_ADDRESS)
-#  include "no_unique_address.hpp"
+#ifdef __cpp_lib_span
+#include "span"
```
mhoemmen: Standard Library headers need `<...>`:
```suggestion
#include <span>
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,505 +13,602 @@
...
 #endif
-
-#include <array>
-#include <cstddef>
+#include "array"
```
mhoemmen:
```suggestion
#include <array>
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,505 +13,602 @@
...
 };
 
-template<>
-struct _count_dynamic_extents<> {
-  static constexpr size_t val = 0;
+// End the recursion
+template <size_t R, class T, T FirstExt>
+struct static_array_impl<R, T, FirstExt> {
+  MDSPAN_INLINE_FUNCTION
+  constexpr static T get(int) { return FirstExt; }
```
mhoemmen:
```suggestion
  constexpr static T get(size_t) { return FirstExt; }
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,505 +13,602 @@
...
->
-_check_compatible_extents(
-  std::true_type, std::integer_sequence<size_t, Extents...>, std::integer_sequence<size_t, OtherExtents...>
-) noexcept { return { }; }
-
-struct __extents_tag { };
+// Don't start recursion if size 0
+template <class T> struct static_array_impl<0, T> {
+  MDSPAN_INLINE_FUNCTION
+  constexpr static T get(int) { return T(); }
```
mhoemmen:
```suggestion
  constexpr static T get(size_t) { return T(); }
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,505 +13,602 @@
...
 public:
+  using value_type = T;
 
-  using rank_type = size_t;
-  using index_type = ThisIndexType;
-  static_assert(std::is_integral<index_type>::value && !std::is_same<index_type, bool>::value,
-                "std::extents requires a signed or unsigned integer as index_type parameter");
-  using size_type = make_unsigned_t<index_type>;
+  MDSPAN_INLINE_FUNCTION
+  constexpr T operator[](int r) const { return impl_t::get(r); }
```
mhoemmen: Would you consider either making these `int` parameters `size_t` for consistency with `static_array_impl`, or using `int` throughout?

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,505 +13,602 @@
...
+};
+
+template <size_t R, size_t FirstVal>
+struct index_sequence_scan_impl<R, FirstVal> {
+#if defined(__NVCC__) || defined(__NVCOMPILER)
+  // NVCC warns about pointless comparison with 0 for R==0 and r being const
+  // evaluatable and also 0.
+  MDSPAN_INLINE_FUNCTION
+  constexpr static size_t get(size_t r) {
+    return static_cast<int64_t>(R) > static_cast<int64_t>(r) ? FirstVal : 0;
```
mhoemmen, line 149: Would you consider rephrasing the comparison to avoid the warning?  It's something that could come up with other compilers, given that `R` is unsigned and definitely known at compile time.
```suggestion
    return R == 0 || R <= r ? size_t(0) : FirstVal;
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,505 +13,602 @@
...
-  }
+  constexpr const T &operator[](size_t r) const { return vals[r]; }
+};
 
-  template <class OtherIndexType, size_t... OtherExtents, size_t... Idxs>
+template <class T> struct possibly_empty_array<T, 0> {
   MDSPAN_INLINE_FUNCTION
-  constexpr bool _not_eq_impl(std::experimental::extents<OtherIndexType, OtherExtents...>, false_type, index_sequence<Idxs...>) const noexcept { return true; }
-  template <class OtherIndexType, size_t... OtherExtents, size_t... Idxs>
+  constexpr T operator[](size_t) { return T(); }
```
mhoemmen, line 180: Do we actually need the empty case to implement `operator[]`?

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,505 +13,602 @@
...
-// public interface:
 public:
-  /* Defined above for use in the private code
-  using rank_type = size_t;
-  using index_type = ThisIndexType;
-  */
-
-  MDSPAN_INLINE_FUNCTION
-  static constexpr rank_type rank() noexcept { return sizeof...(Extents); }
+  // static mapping of indicies to the position in the dynamic values array
```
mhoemmen:
```suggestion
  // static mapping of indices to the position in the dynamic values array
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,505 +13,602 @@
...
+  using dyn_map_t = index_sequence_scan_impl<0, size_t(Values == dyn_tag)...>;
+
+  // two types for static and dynamic values
+  using value_type = TDynamic;
+  using static_value_type = TStatic;
+  // tag value indiciating dynamic value
```
mhoemmen:
```suggestion
  // tag value indicating dynamic value
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,505 +13,602 @@
...
+  constexpr static static_value_type tag_value = dyn_tag;
+
+  constexpr maybe_static_array() = default;
+
+  // constructor for all static values
+  // TODO: add precondition check?
+  MDSPAN_TEMPLATE_REQUIRES(class... Vals,
```
mhoemmen, line 224: Do we also need to add to the constraints that TStatic is constructible from each of the Vals?

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,505 +13,602 @@
...
+                           /* requires */ ((m_size_dynamic == 0) &&
+                                           (sizeof...(Vals) > 0)))
   MDSPAN_INLINE_FUNCTION
-  static constexpr rank_type rank_dynamic() noexcept { return _MDSPAN_FOLD_PLUS_RIGHT((rank_type(Extents == dynamic_extent)), /* + ... + */ 0); }
+  constexpr maybe_static_array(Vals...) : m_dyn_vals{} {}
 
-  //--------------------------------------------------------------------------------
-  // Constructors, Destructors, and Assignment
+  // constructors from dynamic values only
+  MDSPAN_TEMPLATE_REQUIRES(class... DynVals,
```
mhoemmen, line 231: Do we also need to add to the constraints that TDynamic is constructible from each of the DynVals?

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,505 +13,602 @@
...
-     * either
-     *   - sizeof...(OtherExtents) is zero, or
-     *   - other.extent(r) is a representable value of type index_type for all rank index r of other
-     */
+  MDSPAN_TEMPLATE_REQUIRES(class T, size_t N,
+                           /* requires */ (N == m_size_dynamic && N > 0))
+  MDSPAN_INLINE_FUNCTION
+  constexpr maybe_static_array(const std::array<T, N> &vals) {
+    for (size_t r = 0; r < N; r++)
+      m_dyn_vals[r] = static_cast<TDynamic>(vals[r]);
```
mhoemmen, line 245: Should we consider instead constructing `m_dyn_vals` from `vals` by doing an `index_sequence` trick to build the `array`?

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,505 +13,602 @@
...
-    : __storage_{
-#else
-    : __base_t(__base_t{typename __base_t::__stored_type{
+  constexpr maybe_static_array(const std::array<T, N> &) : m_dyn_vals{} {}
+
+#ifdef __cpp_lib_span
+  MDSPAN_TEMPLATE_REQUIRES(class T, size_t N,
+                           /* requires */ (N == m_size_dynamic))
+  MDSPAN_INLINE_FUNCTION
+  constexpr maybe_static_array(const std::span<T, N> &vals) {
```
mhoemmen, line 257: Should we consider constructing `m_dyn_vals` from the `span` using an `index_sequence` trick (see above)?

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,505 +13,602 @@
...
-    // Can't do assert here since that breaks true constexpr ness
-    // assert(n<rank());
-    return _static_extent_impl(n, std::make_integer_sequence<size_t, sizeof...(Extents)>{});
+  constexpr TDynamic value(int r) const {
+    TStatic static_val = static_vals_t::get(r);
+    return static_val == dynamic_extent ? m_dyn_vals[dyn_map_t::get(r)]
+                                        : static_cast<TDynamic>(static_val);
   }
+  MDSPAN_INLINE_FUNCTION
+  constexpr TDynamic operator[](int r) const { return value(r); }
```
mhoemmen: Please see above discussion about `int` vs. `size_t`; thanks!

mhoemmen (commented):
This is a lovely improvement -- it's much easier to read!  Thank you for doing this!

My only concern is minor, that functions like `get` sometimes take `int` and sometimes take `size_t`.  This could lead to conversion warnings.

include/experimental/__p0009_bits/extents.hpp
crtrott: Yes because otherwise simple code like:
```
for(int i=0; i<N; i++) sum += a[i];
```

Won't compile. Technically its just a precondition violation to call this thing ever ...

include/experimental/__p0009_bits/extents.hpp
crtrott: definitely not a constraint. We can add a static_assert though in the class. Doing that.

crtrott: Actually misread you. We could but for now I just rely on upstream ones. And I think it might not be a good idea to do these compile time checks at every level to help with compile time. We do check at the extents class level. So we don't need to check at this implementation detail level again.

include/experimental/__p0009_bits/extents.hpp
crtrott: Same as above.

include/experimental/__p0009_bits/extents.hpp
crtrott: not sure that its worth it. N is compile time known after all.

include/experimental/__p0009_bits/extents.hpp
crtrott: Same as above.

CMakeLists.txt
```diff
@@ -194,10 +194,10 @@ if(MDSPAN_ENABLE_EXAMPLES)
 endif()
 
 if(MDSPAN_ENABLE_BENCHMARKS)
-  if(NOT MDSPAN_CXX_STANDARD STREQUAL "14")
+  if(NOT MDSPAN_CXX_STANDARD STREQUAL "14" AND NOT MDSPAN_CXX_STANDARD STREQUAL "23")
     add_subdirectory(benchmarks)
   else()
-    MESSAGE(FATAL_ERROR "Benchmarks are not available in C++14 mode. Turn MDSPAN_ENABLE_BENCHMARKS OFF or use C++17 or newer.")
+    MESSAGE(FATAL_ERROR "Benchmarks are not available in C++14 or C++23 mode. Turn MDSPAN_ENABLE_BENCHMARKS OFF or use C++17/20 or newer.")
```
nmm0:
```suggestion
    MESSAGE(FATAL_ERROR "Benchmarks are not available in C++14 or C++23 mode. Turn MDSPAN_ENABLE_BENCHMARKS OFF or use C++17/20.")
```

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -272,7 +239,11 @@ struct layout_stride {
 #else
       : __base_t(__base_t{__member_pair_t(
 #endif
+#ifndef _MDSPAN_NEW_EXTENT2S
```
dalg24: What is this macro?

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,612 @@
...
+template <class T> struct static_array_impl<0, T> {
+  MDSPAN_INLINE_FUNCTION
+  constexpr static T get(size_t) { return T(); }
+  template <size_t> MDSPAN_INLINE_FUNCTION constexpr static T get() {
+    return T();
+  }
+};
+
+// Static array, provides get<r>(), get(r) and operator[r]
+template <class T, T... Values> struct static_array {
```
nmm0: Can just inherit from `static_array_impl<0, T, Values...>`

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,612 @@
...
+  constexpr static size_t m_size_dynamic =
+      _MDSPAN_FOLD_PLUS_RIGHT((Values == dyn_tag), 0);
+
+  // Dynamic values member
+  _MDSPAN_NO_UNIQUE_ADDRESS possibly_empty_array<TDynamic, m_size_dynamic>
+      m_dyn_vals;
+
+public:
+  // static mapping of indices to the position in the dynamic values array
+  using dyn_map_t = index_sequence_scan_impl<0, size_t(Values == dyn_tag)...>;
```
nmm0:
```suggestion
  using dyn_map_t = index_sequence_scan_impl<0, static_cast<size_t>(Values == dyn_tag)...>;
```

Also, should this be public?

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,612 @@
...
+template <size_t R, size_t FirstVal, size_t... Values>
+struct index_sequence_scan_impl<R, FirstVal, Values...> {
+  MDSPAN_INLINE_FUNCTION
+  constexpr static size_t get(size_t r) {
+    if (r > R)
+      return FirstVal + index_sequence_scan_impl<R + 1, Values...>::get(r);
+    else
+      return 0;
+  }
+  template <size_t r> MDSPAN_INLINE_FUNCTION constexpr static size_t get() {
```
dalg24: This overload is never used.

dalg24 (commented):
I don't love it but it is mostly fine

dalg24 (approved):

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,612 @@
...
-  */
+  static_assert(is_convertible<TStatic, TDynamic>::value, "maybe_static_array: TStatic must be convertible to TDynamic");
+  static_assert(is_convertible<TDynamic, TStatic>::value, "maybe_static_array: TDynamic must be convertible to TStatic");
 
+private:
+  // Static values member
+  using static_vals_t = static_array<TStatic, Values...>;
+  constexpr static size_t m_size = sizeof...(Values);
+  constexpr static size_t m_size_dynamic =
+      _MDSPAN_FOLD_PLUS_RIGHT((Values == dyn_tag), 0);
```
nmm0, line 204: These can be moved into the respective functions

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,612 @@
...
-#endif
+  MDSPAN_TEMPLATE_REQUIRES(class T, size_t N,
+                           /* requires */ (N == m_size_dynamic && N == 0))
   MDSPAN_INLINE_FUNCTION
-  explicit constexpr extents(Integral... exts) noexcept
-#if defined(_MDSPAN_USE_ATTRIBUTE_NO_UNIQUE_ADDRESS)
-    : __storage_{
-#else
-    : __base_t(__base_t{typename __base_t::__stored_type{
+  constexpr maybe_static_array(const std::array<T, N> &) : m_dyn_vals{} {}
```
nmm0, line 251:
```suggestion
  MDSPAN_TEMPLATE_REQUIRES(class T,
                           /* requires */ (m_size_dynamic == 0))
  MDSPAN_INLINE_FUNCTION
  constexpr maybe_static_array(const std::array<T, 0> &) : m_dyn_vals{} {}
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,612 @@
...
+
+  // constructors from all values
+  MDSPAN_TEMPLATE_REQUIRES(class... DynVals,
+                           /* requires */ (sizeof...(DynVals) !=
+                                               m_size_dynamic &&
+                                           m_size_dynamic > 0))
+  MDSPAN_INLINE_FUNCTION
+  constexpr maybe_static_array(DynVals... vals) {
+    static_assert((sizeof...(DynVals) == m_size), "Invalid number of values.");
+    TDynamic values[m_size]{static_cast<TDynamic>(vals)...};
```
nmm0, line 271: I think I would prefer `std::array` here

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,612 @@
...
+      class T, size_t N,
+      /* requires */ (N != m_size_dynamic && m_size_dynamic > 0))
   MDSPAN_INLINE_FUNCTION
-  constexpr
-  extents(std::array<IndexType, N> const& exts) noexcept
-#if defined(_MDSPAN_USE_ATTRIBUTE_NO_UNIQUE_ADDRESS)
-    : __storage_{
-#else
-    : __base_t(__base_t{typename __base_t::__stored_type{
+  constexpr maybe_static_array(const std::array<T, N> &vals) {
```
nmm0, line 290: Don't need `std` here

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,612 @@
...
+    assert(N == m_size);
 #endif
-      std::conditional_t<N==rank_dynamic(),
-        detail::__construct_psa_from_dynamic_exts_array_tag_t<0>,
-        detail::__construct_psa_from_all_exts_array_tag_t>(),
-      exts
-#if defined(_MDSPAN_USE_ATTRIBUTE_NO_UNIQUE_ADDRESS)
+    for (size_t r = 0; r < m_size; r++) {
+      TStatic static_val = static_vals_t::get(r);
+      if (static_val == dynamic_extent) {
```
nmm0:
```suggestion
      if (static_val == dyn_tag) {
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,612 @@
...
+    assert(N == m_size);
 #endif
-      std::conditional_t<N==rank_dynamic(),
-        detail::__construct_psa_from_dynamic_exts_array_tag_t<0>,
-        detail::__construct_psa_from_all_exts_array_tag_t>(),
-      std::array<IndexType,N>{exts}
-#if defined(_MDSPAN_USE_ATTRIBUTE_NO_UNIQUE_ADDRESS)
+    for (size_t r = 0; r < m_size; r++) {
+      TStatic static_val = static_vals_t::get(r);
+      if (static_val == dynamic_extent) {
```
nmm0:
```suggestion
      if (static_val == dyn_tag) {
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,612 @@
...
+                           /* requires */ (sizeof...(DynVals) !=
+                                               m_size_dynamic &&
+                                           m_size_dynamic > 0))
+  MDSPAN_INLINE_FUNCTION
+  constexpr maybe_static_array(DynVals... vals) {
+    static_assert((sizeof...(DynVals) == m_size), "Invalid number of values.");
+    TDynamic values[m_size]{static_cast<TDynamic>(vals)...};
+    for (size_t r = 0; r < m_size; r++) {
+      TStatic static_val = static_vals_t::get(r);
+      if (static_val == dynamic_extent) {
```
nmm0:
```suggestion
      if (static_val == dyn_tag) {
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,612 @@
...
 
   MDSPAN_INLINE_FUNCTION
-  static constexpr
-  size_t static_extent(size_t n) noexcept {
-    // Can't do assert here since that breaks true constexpr ness
-    // assert(n<rank());
-    return _static_extent_impl(n, std::make_integer_sequence<size_t, sizeof...(Extents)>{});
+  constexpr TDynamic value(size_t r) const {
+    TStatic static_val = static_vals_t::get(r);
+    return static_val == dynamic_extent ? m_dyn_vals[dyn_map_t::get(r)]
```
nmm0:
```suggestion
    return static_val == dyn_tag ? m_dyn_vals[dyn_map_t::get(r)]
```

CMakeLists.txt
mhoemmen:
```suggestion
    MESSAGE(FATAL_ERROR "Benchmarks are not available in C++14 or C++23 mode. Turn MDSPAN_ENABLE_BENCHMARKS OFF or use C++17 or C++20.")
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,612 @@
...
-  static constexpr size_t val = (E==dynamic_extent?1:0) + _count_dynamic_extents<Extents...>::val;
-};
-
-template<>
-struct _count_dynamic_extents<> {
-  static constexpr size_t val = 0;
-};
-
+// Function used to check compatibility of extents in converting constructor
+// can't be private member functions for some reason.
```
mhoemmen:
```suggestion
// can't be a private member function for some reason.
```

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,612 @@
...
+struct static_array_impl<R, T, FirstExt> {
+  MDSPAN_INLINE_FUNCTION
+  constexpr static T get(size_t) { return FirstExt; }
+  template <size_t> MDSPAN_INLINE_FUNCTION constexpr static T get() {
+    return FirstExt;
+  }
+};
+
+// Don't start recursion if size 0
+template <class T> struct static_array_impl<0, T> {
```
nmm0, line 101: I'm not sure about this, I know it's impl code and we are returning some value that won't be used, but I would rather specialize `maybe_static_array` for size 0. Then just don't define the get functions

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,612 @@
...
+  constexpr static size_t get(size_t r) {
+    return static_cast<int64_t>(R) > static_cast<int64_t>(r) ? FirstVal : 0;
+  }
+#else
+  MDSPAN_INLINE_FUNCTION
+  constexpr static size_t get(size_t r) { return R > r ? FirstVal : 0; }
+#endif
+  template <size_t r> MDSPAN_INLINE_FUNCTION constexpr static size_t get() {
+#if MDSPAN_HAS_CXX_17
+    if constexpr (r > R) {
```
nmm0: Does this situation actually come up in the code? The return value of `FirstVal` doesn't reflect the stated result of this operation. Wouldn't this only occur on an out of bounds access?

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,612 @@
...
+// ------------ possibly_empty_array  -------------------------------
+// ------------------------------------------------------------------
 
-  template <class OtherIndexType, size_t... OtherExtents, size_t... Idxs>
+// array like class which provides get function and operator [], and
+// has a specialization for the size 0 case.
+// This is needed to make the maybe_static_array be truly empty, for
+// all static values.
+
+template <class T, size_t N> struct possibly_empty_array {
```
nmm0, line 170: We can just use `std::array` instead unless you want to allow out of bounds access

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,612 @@
...
-  ) const noexcept {
-    return _MDSPAN_FOLD_AND(
-      (__storage().template __get_n<Idxs>() == other.__storage().template __get_n<Idxs>()) /* && ... */
-    );
-  }
+  constexpr const T &operator[](size_t r) const { return vals[r]; }
+};
 
-  template <class OtherIndexType, size_t... OtherExtents, size_t... Idxs>
+template <class T> struct possibly_empty_array<T, 0> {
```
nmm0, line 178: In this case I would rather specialize `maybe_static_array` for size 0

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,612 @@
...
+    return FirstExt;
+  }
+};
+
+// Don't start recursion if size 0
+template <class T> struct static_array_impl<0, T> {
+  MDSPAN_INLINE_FUNCTION
+  constexpr static T get(size_t) { return T(); }
+  template <size_t> MDSPAN_INLINE_FUNCTION constexpr static T get() {
+    return T();
```
mhoemmen, line 105: Calling `get` on an empty array is definitely a precondition violation.  Thus, would you consider using `__builtin_unreachable()` with GCC, or `std::unreachable` with C++23, with functions like this?  This works fine in practice with `constexpr` functions.

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,612 @@
...
+  // constructor for all static values
+  // TODO: add precondition check?
+  MDSPAN_TEMPLATE_REQUIRES(class... Vals,
+                           /* requires */ ((m_size_dynamic == 0) &&
+                                           (sizeof...(Vals) > 0)))
   MDSPAN_INLINE_FUNCTION
-  static constexpr rank_type rank() noexcept { return sizeof...(Extents); }
-  MDSPAN_INLINE_FUNCTION
-  static constexpr rank_type rank_dynamic() noexcept { return _MDSPAN_FOLD_PLUS_RIGHT((rank_type(Extents == dynamic_extent)), /* + ... + */ 0); }
+  constexpr maybe_static_array(Vals...) : m_dyn_vals{} {}
```
mhoemmen, line 259: Could this constructor case be `=default`ed?

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,612 @@
...
+      class T, size_t N,
+      /* requires */ (N != m_size_dynamic && m_size_dynamic > 0))
   MDSPAN_INLINE_FUNCTION
-  constexpr
-  extents(std::span<IndexType, N> exts) noexcept
-#if defined(_MDSPAN_USE_ATTRIBUTE_NO_UNIQUE_ADDRESS)
-    : __storage_{
-#else
-    : __base_t(__base_t{typename __base_t::__stored_type{
+  constexpr maybe_static_array(const std::span<T, N> &vals) {
```
mhoemmen, line 316: Should we consider passing `span` by value instead of by const reference?

nmm0 (changes requested):
I think my biggest issue is returning values if the size of `maybe_static_array` is 0. I would rather specialize and maybe even assert if the get functions are accessed

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,612 @@
...
 
-  //--------------------------------------------------------------------------------
+// Class to describe the extents of a multi dimensional array.
+// Used by mdspan, mdarray and layout mappings.
+// See ISO C++ standard [mdspan.extents]
 
-  template<class OtherIndexType, size_t... RHS>
+template <class IndexType, size_t... Extents> class extents {
+public:
+  // typedefs for integral types used
```
mhoemmen, line 374: Should we consider checking the Mandate that `IndexType` is a signed or unsigned integral type?  That would effectively make the code self-documenting, so that the comment would become superfluous.

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,612 @@
...
-  template <size_t N, size_t Default=dynamic_extent>
+public:
+
+  // Converting constructor from other extents specializations
+  MDSPAN_TEMPLATE_REQUIRES(
+      class OtherIndexType, size_t... OtherExtents,
+      /* requires */
+      (
+          /* multi-stage check to protect from invalid pack expansion when sizes
+             don't match? */
```
mhoemmen, line 501: Do we need the question mark here?

mhoemmen (approved):
Hi Christian!  This is much more concise; I like it!  I just made a few optional suggestions.

include/experimental/__p0009_bits/extents.hpp
crtrott: std::array doesn't work on device everywhere.

nmm0: I see -- though I still don't like the complexity of dealing with the out of bounds access for the size 0 case. I think this should just be replaced with `T[N]` then

include/experimental/__p0009_bits/extents.hpp
crtrott: If we don't specialize here we need to specialize both static_array and maybe_static_array

include/experimental/__p0009_bits/extents.hpp
crtrott: I really don't want to add ifdefs and builtins inside this implementation only code.

include/experimental/__p0009_bits/extents.hpp
crtrott: I can try this.

include/experimental/__p0009_bits/extents.hpp
crtrott: I still kinda would go for consistency here across the different things which looks similar?

include/experimental/__p0009_bits/extents.hpp
nmm0: nvm I didn't think of device code

include/experimental/__p0009_bits/extents.hpp
crtrott: this is the wrong way around? shouldn't it be R>r?

crtrott: But as Damien observed above this overload is not used?

include/experimental/__p0009_bits/extents.hpp
crtrott: but its not maybe_static_array for size 0, its maybe_static_array with dynamic size 0 we would need to specialize.

crtrott: which means we would need to add dummy parameters to the template arguments of maybe_static_array

crtrott: Oh and to make that clear, I think that would be worse.

include/experimental/__p0009_bits/extents.hpp
crtrott: Yeah I guess I can make this private now (at some point I used it inside of extents itself)

include/experimental/__p0009_bits/extents.hpp
crtrott: Doesn't work because of std::array not working on device.

include/experimental/__p0009_bits/extents.hpp
crtrott: std::array doesn't work on device.

include/experimental/__p0009_bits/extents.hpp
crtrott: ah yeah ...

include/experimental/__p0009_bits/extents.hpp
crtrott: yes we need it in the future, because I want to make namespace configurable.

include/experimental/__p0009_bits/extents.hpp
crtrott: jup

include/experimental/__p0009_bits/extents.hpp
crtrott: not sure.

include/experimental/__p0009_bits/extents.hpp
crtrott: Yeah I can do that.

include/experimental/__p0009_bits/extents.hpp
crtrott: This question mark at least to me indicates that this is a weird unexpected problem

include/experimental/__p0009_bits/layout_stride.hpp
crtrott: leftover from when you could switch in the PR between the old and new implementation, gonna remove.

include/experimental/__p0009_bits/extents.hpp
crtrott: Unfortunately that doesn't do proper empty class optimization, and we won't pass the size test of extents.

include/experimental/__p0009_bits/extents.hpp
crtrott: ok

crtrott: actually m_size_dynamic is used right below here, I guess if I moved that member to the end so the constexpr accessor functions are already defined? But I don't see a big benefit since they are anyway private.

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -22,6 +22,8 @@
 #endif
 #include <array>
 
+#include<inttypes.h>
```
mhoemmen:
```suggestion
#include <cinttypes>
```

include/experimental/__p0009_bits/extents.hpp
nmm0: Yeah it seems like it would be a bit more complicated let's keep as is

include/experimental/__p0009_bits/extents.hpp
nmm0: I think we should at least assert since it will be called on out of bounds access and `static_assert` where possible

include/experimental/__p0009_bits/extents.hpp
nmm0: Ok, makes sense, I'm convinced

include/experimental/__p0009_bits/extents.hpp
nmm0: Ok, makes sense

include/experimental/__p0009_bits/extents.hpp
nmm0: It's just less code. I'm not dead set on this

include/experimental/__p0009_bits/extents.hpp
nmm0: Oh weird, my diff got messed up idk why it grabbed some of the old code. The only change here was to simplify
```cpp
MDSPAN_TEMPLATE_REQUIRES(class T, size_t N,
                           /* requires */ (N == m_size_dynamic && N == 0))
MDSPAN_INLINE_FUNCTION
constexpr maybe_static_array(const std::array<T, N> &) : m_dyn_vals{} {}
```

to 

```cpp
MDSPAN_TEMPLATE_REQUIRES(class T,
                           /* requires */ (m_size_dynamic == 0))
MDSPAN_INLINE_FUNCTION
constexpr maybe_static_array(const std::array<T, 0> &) : m_dyn_vals{} {}
```

-- you are already using array here so this isn't introducing it in a new place

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -13,511 +13,583 @@
...
+  public static_array_impl<0, T, Values...>  {
 
-template <class ThisIndexType, size_t... Extents>
-class extents
-#if !defined(_MDSPAN_USE_ATTRIBUTE_NO_UNIQUE_ADDRESS)
-  : private detail::__no_unique_address_emulation<
-      detail::__partially_static_sizes_tagged<detail::__extents_tag, ThisIndexType , size_t, Extents...>>
-#endif
-{
 public:
```
nmm0, line 113: Very minor point, but both `public`s here are redundant

nmm0 (approved):
I think overall I would want the out of bounds checks to be asserted/static_asserted when able but otherwise looks good!

mhoemmen:
@nmm0 wrote:
> I think we should at least assert since it will be called on out of bounds access and `static_assert` where possible

regarding

```c++
 template <size_t> MDSPAN_INLINE_FUNCTION constexpr static T get() {
    return T();
}
```

another approach would be to use the C++23 feature `std::unreachable()`, or one of the compiler-specific equivalents, like GCC's `__builtin_unreachable()`.

include/experimental/__p0009_bits/extents.hpp
crtrott: The above thing is not the same (where you say m_size_dynamic==0). Think N =1 and m_size_dynamic = 0, your case would take it, whats there right now doesn.t

PR #214, opened 2022-11-22
Make operator implementation standard conformant

### LINKED ISSUE #186: mdspan::operator() is inconsistent with operator[]; both are inconsistent with the spec
Based on my comment here: https://github.com/kokkos/mdspan/issues/109#issuecomment-1235817012

The spec voted into the C++ draft [says](http://eel.is/c++draft/mdspan.mdspan.members) that `mdspan::operator[]` has effects equivalent to `acc_.access(ptr_, map_(static_cast<index_type>(std::move(indices))...));`.

There are two issues with the current mdspan implementation.  First, the backwards compatibility parentheses indexing operator `mdspan::operator()` is inconsistent with `mdspan::operator[]`.  `operator()` does _not_ cast the indices, while `operator[]` does.

* `operator()`: https://github.com/kokkos/mdspan/blob/d58ddaaa1869aa5ad8469c3b02abe025efab52a4/include/experimental/__p0009_bits/mdspan.hpp#L310
* `operator[]`: https://github.com/kokkos/mdspan/blob/d58ddaaa1869aa5ad8469c3b02abe025efab52a4/include/experimental/__p0009_bits/mdspan.hpp#L250

Second, neither of these two operators `std::move` the indices into the layout mapping's `operator()`.  This matters only for index types that are not integral and have nontrivial state.  (An example of a nonintegral index type would be a "tagged index type" that can be useful for avoiding confusion between different loop indices.  However, these are usually just wrappers around an integer.)  Nevertheless, it's not hard for us to fix this.

crtrott:
applies to mdspan and mdarray.

Fixes #186

### DIFF
--- a/include/experimental/__p0009_bits/mdspan.hpp
+++ b/include/experimental/__p0009_bits/mdspan.hpp
@@ -247,7 +247,7 @@ public:
   MDSPAN_FORCE_INLINE_FUNCTION
   constexpr reference operator[](SizeTypes... indices) const
   {
-    return __accessor_ref().access(__ptr_ref(), __mapping_ref()(index_type(indices)...));
+    return __accessor_ref().access(__ptr_ref(), __mapping_ref()(static_cast<index_type>(std::move(indices))...));
   }
   #endif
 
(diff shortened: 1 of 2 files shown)


nmm0 (commented):
Looks good to me

nmm0 (approved):

mhoemmen (approved):

Issue #193, opened 2022-09-30
layout_stride mapping constructor taking (extents_type, span) doesn't compile


mhoemmen:
@youyu3 discovered that the `layout_stride::mapping(const extents_type&, span<OtherIndexType, rank()>)` constructor doesn't compile.  Here is an example:

https://godbolt.org/z/Tf57xdrGq

```c++
    using extents_type = stdex::dextents<int, 2>;
    extents_type ex{2, 2};

    // this one works fine
    stdex::layout_stride::mapping map0{ex, std::array{3, 3}};
 
    std::array<int, 2> strides{3, 3};
    std::span<int, 2> strides_span{strides.begin(), strides.end()};

    // Neither of these compile 
    stdex::layout_stride::mapping map1{ex, strides_span};
    stdex::layout_stride::mapping<extents_type> map2{ex, strides_span};
```

@youyu3 suggests a fix: add a `__deduction_workaround::fill_strides` overload taking `span<IntegerType, extents_type::rank()>`, like the existing one that takes `array<IntegerType, extents_type::rank()>` ( https://github.com/kokkos/mdspan/blob/760060059fb746018a9849234e02dc9bf003861b/include/experimental/__p0009_bits/layout_stride.hpp#L197 ):

```c++
template<class IntegralType>
MDSPAN_INLINE_FUNCTION
static constexpr const __strides_storage_t
fill_strides(const span<IntegralType, extents_type::rank()>& s) {
  return __strides_storage_t{static_cast<index_type>(s[Idxs])...};
}
```

PR #215, opened 2022-11-22
Fix layout_stride construction from extents and span

crtrott:
Addresses #193  and adds test.

### DIFF
--- a/include/experimental/__p0009_bits/layout_stride.hpp
+++ b/include/experimental/__p0009_bits/layout_stride.hpp
@@ -56,6 +56,9 @@
 #include <algorithm>
 #include <numeric>
 #include <array>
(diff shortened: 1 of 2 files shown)


nmm0 (approved):
Looks good to me!

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -198,6 +201,14 @@ struct layout_stride {
         return __strides_storage_t{static_cast<index_type>(s[Idxs])...};
       }
 
+#ifdef __cpp_lib_span
```
mhoemmen, line 204: @youyu3 If I remember right, I think you reported the bug that this fixes.

mhoemmen (approved):

PR #216, opened 2022-11-23
Add static assert for using integral type for extents

### LINKED ISSUE #188: extents does not enforce Mandate that IndexType is integral
Due to a recent build issue in a different project, I accidentally constructed an `extents` type where the first template parameter `IndexType` was an `extents` specialization, instead of an integral type.  Nevertheless, the code compiled, even though the spec Mandates that "`IndexType` is a signed or unsigned integer type."  

In a brief Compiler Explorer example, I got more complicated build errors: https://godbolt.org/z/eszjMcW3n

In any case, it should be straightforward to add a `static_assert` for is an integer but not `bool` or `{signed,unsigned,} char`.

crtrott:
Fixes #188

### DIFF
--- a/include/experimental/__p0009_bits/extents.hpp
+++ b/include/experimental/__p0009_bits/extents.hpp
@@ -108,6 +108,8 @@ public:
 
   using rank_type = size_t;
   using index_type = ThisIndexType;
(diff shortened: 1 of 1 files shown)


nmm0 (approved):
Looks good!

mhoemmen (approved):
@youyu3 Check this out

Issue #39, opened 2021-04-10
constructing extents from array fails at compile time


jbigot:
My attempts to use the following constructor failed at build time with GCC-10.2.1 configured for C++-17 support in cmake:
```C++
template<ptrdiff_t... Extents>
class extents {
  //...
  template<class IndexType>
    constexpr extents(const array<IndexType, rank_dynamic()>&) noexcept;
};
```
The error is:
```
In file included from /mdspan_test/vendor/mdspan/include/experimental/__p0009_bits/fixed_layout_impl.hpp:50,
                 from /mdspan_test/vendor/mdspan/include/experimental/__p0009_bits/layout_right.hpp:47,
                 from /mdspan_test/vendor/mdspan/include/experimental/__p0009_bits/basic_mdspan.hpp:48,
                 from /mdspan_test/vendor/mdspan/include/experimental/mdspan:48,
                 from /mdspan_test/mdspan_test.cpp:2:
/mdspan_test/vendor/mdspan/include/experimental/__p0009_bits/extents.hpp: In instantiation of ‘constexpr std::experimental::extents<Extents>::extents(const std::array<IndexType, std::experimental::extents<Extents>::rank_dynamic()>&) [with IndexType = long unsigned int; typename std::enable_if<is_convertible_v<IndexType, long int>, int>::type <anonymous> = 0; long int ...Extents = {-1, -1, -1}]’:
/mdspan_test/mdspan_test.cpp:11:85:   required from here
/mdspan_test/vendor/mdspan/include/experimental/__p0009_bits/extents.hpp:208:69: error: no type named ‘stored_type’ in ‘struct std::experimental::detail::__no_unique_address_emulation<std::experimental::detail::__partially_static_sizes_tagged<std::experimental::detail::__extents_tag, -1, -1, -1>, 0, void>’
  208 |         detail::__construct_psa_from_dynamic_values_tag_t<>{}, dyn}})
      |                                                                     ^
```

More info including a complete example and the build log can be found at https://github.com/jbigot/mdspan_test

Did I make a mistake in using this constructor?

jbigot:
Looking at the open PRs, this is already fixed by #36 . Would be nice to have it merged in master

crtrott:
there is a related issue that it doesn't construct from a std::array of any integer type ...

crtrott:
i think this is fixed. Added test in #217

PR #217, opened 2022-11-23
Check constructibility of extents from array and span

crtrott:
Make sure we have test for #39


nmm0 (approved):
👍

mhoemmen (approved):
I think @youyu3 wrote a TON of tests like these.  Back-porting could make sense.  The only issue is the difference in test framework.

PR #223, opened 2022-12-29
Added new submdspan implementation

crtrott:
This uses ADL customization point design.

It doesn't use the backward compatibility stuff yet to improve readability for an initial review, i.e. it will only work on host with C++20 and C++23.

It also doesn't implement yet correct stride calculation when using strided_index_range

### DIFF
--- a/include/experimental/__p2630_bits/submdspan.hpp
+++ b/include/experimental/__p2630_bits/submdspan.hpp
@@ -9,9 +9,10 @@ constexpr auto
 submdspan(const mdspan<ElementType, Extents, LayoutPolicy, AccessorPolicy> &src,
           SliceSpecifiers... slices) {
   const auto sub_mapping_offset = submdspan_mapping(src.mapping(), slices...);
-  return mdspan(src.accessor().offset(src.data_handle(), sub_mapping_offset.offset),
-                sub_mapping_offset.mapping,
-                typename AccessorPolicy::offset_policy(src.accessor()));
+  return mdspan(
+      src.accessor().offset(src.data_handle(), sub_mapping_offset.offset),
+      sub_mapping_offset.mapping,
+      typename AccessorPolicy::offset_policy(src.accessor()));
 }
 } // namespace experimental
 } // namespace std
--- a/include/experimental/__p2630_bits/submdspan.hpp
+++ b/include/experimental/__p2630_bits/submdspan.hpp
@@ -1,0 +1,17 @@
+#include "submdspan_extents.hpp"
+#include "submdspan_mapping.hpp"
+
+namespace std {
+namespace experimental {
+template <class ElementType, class Extents, class LayoutPolicy,
+          class AccessorPolicy, class... SliceSpecifiers>
+constexpr auto
+submdspan(const mdspan<ElementType, Extents, LayoutPolicy, AccessorPolicy> &src,
+          SliceSpecifiers... slices) {
+  const auto sub_map_offset = submdspan_mapping(src.mapping(), slices...);
+  return mdspan(src.accessor().offset(src.data_handle(), sub_map_offset.offset),
+                sub_map_offset.map,
+                typename AccessorPolicy::offset_policy(src.accessor()));
+}
+} // namespace experimental
+} // namespace std
--- a/include/experimental/__p2630_bits/submdspan.hpp
+++ b/include/experimental/__p2630_bits/submdspan.hpp
@@ -8,9 +8,9 @@ template <class ElementType, class Extents, class LayoutPolicy,
 constexpr auto
 submdspan(const mdspan<ElementType, Extents, LayoutPolicy, AccessorPolicy> &src,
           SliceSpecifiers... slices) {
-  const auto sub_map_offset = submdspan_mapping(src.mapping(), slices...);
-  return mdspan(src.accessor().offset(src.data_handle(), sub_map_offset.offset),
-                sub_map_offset.map,
+  const auto sub_mapping_offset = submdspan_mapping(src.mapping(), slices...);
+  return mdspan(src.accessor().offset(src.data_handle(), sub_mapping_offset.offset),
+                sub_mapping_offset.mapping,
                 typename AccessorPolicy::offset_policy(src.accessor()));
 }
 } // namespace experimental
--- a/include/experimental/__p2630_bits/submdspan_extents.hpp
+++ b/include/experimental/__p2630_bits/submdspan_extents.hpp
@@ -1,0 +1,238 @@
+namespace std {
+namespace experimental {
+
+// Slice Specifier allowing for strides and compile time extent
+template <class OffsetType, class ExtentType, class StrideType>
+struct strided_index_range {
+  using offset_type = OffsetType;
+  using extent_type = ExtentType;
+  using stride_type = StrideType;
+
+  OffsetType offset;
+  ExtentType extent;
+  StrideType stride;
+};
+
+namespace detail {
+
+// Mapping from submapping ranks to srcmapping ranks
+// InvMapRank is gonna be a index_sequence, which we build recursively
+// to contain the mapped indicies.
+template <size_t Counter, class InvMapRank, class... SliceSpecifiers>
+struct inv_map_rank;
+
+// specialization reducing rank by one (i.e. integral slice specifier)
+template <size_t Counter, class Slice, class... SliceSpecifiers,
+          size_t... MapIdxs>
+requires(is_convertible_v<Slice, size_t>) struct inv_map_rank<
+    Counter, index_sequence<MapIdxs...>, Slice, SliceSpecifiers...> {
+  using type = typename inv_map_rank<Counter + 1, index_sequence<MapIdxs...>,
+                                     SliceSpecifiers...>::type;
+};
+
+// specialization for slice specifiers expressing some form of range
+template <size_t Counter, class Slice, class... SliceSpecifiers,
+          size_t... MapIdxs>
+requires(!is_convertible_v<Slice, size_t>) struct inv_map_rank<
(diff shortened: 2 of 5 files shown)


include/experimental/__p2630_bits/submdspan.hpp
```diff
@@ -0,0 +1,17 @@
...
+#include "submdspan_mapping.hpp"
+
+namespace std {
+namespace experimental {
+template <class ElementType, class Extents, class LayoutPolicy,
+          class AccessorPolicy, class... SliceSpecifiers>
+constexpr auto
+submdspan(const mdspan<ElementType, Extents, LayoutPolicy, AccessorPolicy> &src,
+          SliceSpecifiers... slices) {
+  const auto sub_map_offset = submdspan_mapping(src.mapping(), slices...);
```
mhoemmen: Should we consider forwarding the slice specifiers, as in the code below?  It shouldn't make a performance difference for the current slice specifiers (that are all structs of small value types anyway); it's more a specification question.

```suggestion
template <class ElementType, class Extents, class LayoutPolicy,
          class AccessorPolicy, class... SliceSpecifiers>
constexpr auto
submdspan(const mdspan<ElementType, Extents, LayoutPolicy, AccessorPolicy> &src,
          SliceSpecifiers&&... slices) {
  const auto sub_map_offset = submdspan_mapping(src.mapping(), forward<SliceSpecifiers>(slices)...);
```

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -0,0 +1,238 @@
...
+template <size_t Counter, class Slice, class... SliceSpecifiers,
+          size_t... MapIdxs>
+requires(!is_convertible_v<Slice, size_t>) struct inv_map_rank<
+    Counter, index_sequence<MapIdxs...>, Slice, SliceSpecifiers...> {
+  using type =
+      typename inv_map_rank<Counter + 1, index_sequence<MapIdxs..., Counter>,
+                            SliceSpecifiers...>::type;
+};
+
+// final specialziation containing the final index_sequence
```
mhoemmen: A minor spelling issue:
```suggestion
// final specialization containing the final index_sequence
```
Does "final specialization" mean "this is the end of the 'recursion' on `inv_map_rank` specializations"?

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -0,0 +1,238 @@
...
+  OffsetType offset;
+  ExtentType extent;
+  StrideType stride;
+};
+
+namespace detail {
+
+// Mapping from submapping ranks to srcmapping ranks
+// InvMapRank is gonna be a index_sequence, which we build recursively
+// to contain the mapped indicies.
```
mhoemmen:
```suggestion
// InvMapRank is an index_sequence, which we build recursively
// to contain the mapped indices.
```

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -0,0 +1,238 @@
...
+template <size_t Counter, class InvMapRank, class... SliceSpecifiers>
+struct inv_map_rank;
+
+// specialization reducing rank by one (i.e. integral slice specifier)
```
mhoemmen:
```suggestion
// specialization reducing rank by one (i.e., integral slice specifier)
```

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -0,0 +1,238 @@
...
+  return get<0>(i);
+}
+
+template <class OffsetType, class ExtentType, class StrideType>
+constexpr OffsetType
+first_of(const strided_index_range<OffsetType, ExtentType, StrideType> &r) {
+  return r.offset;
+}
+
+// last_of(slice): getting end of slice sepcifier range
```
mhoemmen:
```suggestion
// last_of(slice): getting end of slice specifier range
```

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -0,0 +1,238 @@
...
+template <class Integral0, Integral0 val0, class Integral1, Integral1 val1>
+struct StaticExtent<std::integral_constant<Integral0, val0>,
+                    std::integral_constant<Integral1, val1>> {
+  constexpr static size_t value = val1 - val0;
+};
+
+// creates new extents through recursive calls to next_extent member function
+// next_extent has different overloads for different types of stride specifiers
+template <size_t K, class Extents, size_t... NewExtents>
+struct extents_constructor {
```
mhoemmen, line 176: This is fine.  I think I can see why you chose a C++11-flavored design; it's a pain to have to pass along `K` and `NewExtents...` as function arguments, especially since `NewExtents...` would need to be packed up in a `tuple` first.

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -0,0 +1,238 @@
...
+  template <class Slice, class... SliceSpecifiers>
+  requires(
+      !is_convertible_v<Slice, size_t> &&
+      !is_strided_index_range<Slice>::
+          value) constexpr static auto next_extent(const Extents &ext,
```
mhoemmen: Would you consider reformatting to make this more consistent with the other overloads?

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -0,0 +1,157 @@
+
+namespace std {
+namespace experimental {
+
+//******************************************
+// Return type of submdspan_mapping overloads
+//******************************************
+template <class Mapping> struct mapping_offset {
+  Mapping map;
```
mhoemmen: Would you consider calling this `mapping` instead of `map`, to avoid collision with `std::map`?
```suggestion
  Mapping mapping;
```

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -0,0 +1,157 @@
...
+  constexpr static bool value =
+      (SubRank == 0) ||
+      ((Idx < SubRank - 1
+            ? is_same_v<SliceSpecifiers, full_extent_t>
+            : (Idx == SubRank - 1 ? is_same_v<SliceSpecifiers, full_extent_t> ||
+                                        is_convertible_v<SliceSpecifiers,
+                                                         tuple<size_t, size_t>>
+                                  : true)) &&
+       ...);
+};
```
mhoemmen: The expression for `value` is a bit hard to read, but it's also hard to simplify it, given that all the repeated `bool`s in it (like `is_same_v<SliceSpecifiers, full_extent_t>`) depend on a pack.  Rewriting them into `tuple`s or `array`s somehow makes it worse.

Would you consider changing the term
```c++
Idx == SubRank - 1 ? is_same_v<SliceSpecifiers, full_extent_t> ||
                                        is_convertible_v<SliceSpecifiers,
                                                         tuple<size_t, size_t>>
                                  : true)
```
into the following?
```c++
Idx != SubRank - 1 or
  is_same_v<SliceSpecifiers, full_extent_t> or
  is_convertible_v<SliceSpecifiers, tuple<size_t, size_t>>
```

mhoemmen (approved):

include/experimental/__p2630_bits/submdspan_extents.hpp
crtrott: Jup that was where I was coming from. Its certainly possible to do it but this seemed cleaner to me - though I do understand that different approaches feel cleaner to different people in different ways.

include/experimental/__p2630_bits/submdspan_mapping.hpp
crtrott: yeah should do this.

include/experimental/__p2630_bits/submdspan_mapping.hpp
crtrott: ah that is better :-)

include/experimental/__p2630_bits/submdspan.hpp
crtrott: we could try

crtrott: When I tried this I ran into some trouble with failing tests, i.e. some deduction went haywire, my guess is that all the internal logic also needs to account for && including all the helper functions for matching and counting etc.

include/experimental/__p2630_bits/submdspan_extents.hpp
crtrott: yeah

crtrott: I updated the comment to express this better

include/experimental/__p2630_bits/submdspan_mapping.hpp
crtrott: actually that expression is wrong ...
this thing will return true for integers and what not.

crtrott: I inserted a correct one which is more structured like this instead of nested ?: statements.

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -0,0 +1,174 @@
+#include <tuple>
+namespace std {
```
dalg24, line 2: You forgot the copyright header

include/experimental/__p2630_bits/submdspan.hpp
mhoemmen: Thanks for trying!  I think it's OK to leave this alone for now.  If LWG wants it we can investigate it then.

PR #226, opened 2023-01-06
Fix `make_single_header.py` to satisfy lower version compiler when us…

Ghost-LZW:
Fix `make_single_header.py` to satisfy lower version compiler when useing single_header version mdspan

Format with black

### DIFF
--- a/make_single_header.py
+++ b/make_single_header.py
@@ -7,6 +7,7 @@ from os.path import dirname, join as path_join, abspath, exists
 
 extra_paths = [path_join(dirname(abspath(__file__)), "include")]
 
+
 def find_file(included_name, current_file):
     current_dir = dirname(abspath(current_file))
     for idir in [current_dir] + extra_paths:
@@ -15,7 +16,14 @@ def find_file(included_name, current_file):
             return try_path
     return None
 
-def process_file(file_path, out_lines=[], front_matter_lines=[], back_matter_lines=[], processed_files=[]):
+
+def process_file(
+    file_path,
+    out_lines=[],
+    front_matter_lines=[],
+    back_matter_lines=[],
+    processed_files=[],
+):
     out_lines += "//BEGIN_FILE_INCLUDE: " + file_path + "\n"
     with open(file_path, "r") as f:
         for line in f:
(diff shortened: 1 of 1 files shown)


Ghost-LZW:
> add it to the front matter just to be clean

May break some define condition, eg:

```
#if __has_include(<version>)
#  include <version>
#else
#  include <type_traits>
#  include <utility>
#endif
```

in https://github.com/kokkos/mdspan/blob/stable/include/experimental/__p0009_bits/config.hpp#L22-L27

Ghost-LZW:
https://godbolt.org/z/d9jGjYeE8 Test success under clang7

make_single_header.py
```diff
@@ -26,27 +35,50 @@ def process_file(file_path, out_lines=[], front_matter_lines=[], back_matter_lin
...
+                            processed_files,
+                        )
                     else:
-                        # assume it's a system header; add it to the front matter just to be clean
-                        front_matter_lines += [line]
+                        # assume it's a system header
+                        if inc_name not in system_headers:
+                            out_lines += [line]
+                        else:
+                            system_headers += [inc_name]
```
crtrott: I don't quite understand this. In the else branch aren't you adding inc_name to system_headers again? Since the condition to hit the else branch is that inc_name is already in system_headers?

crtrott (changes requested):
So if I understand this right: previously We moved all the include lines of things without path prefix to the front. Now you want to leave it in place (i.e. just add it to out_lines). I don't understand the logic in this block however:
```
                        # assume it's a system header
                        if inc_name not in system_headers:
                            out_lines += [line]
                        else:
                            system_headers += [inc_name]
```
We start with empty system_headers. So any inc_name which doesn't have a prepend path will just be left in place. It will never hit the else branch because nothing is in system_headers, and the only way that something would be added is if its already in system_headers?

make_single_header.py
Ghost-LZW: Oops, my bad, this block is a mistake. We should just add it to out_lines, without any check. otherwise, there will be conflicted with eg: [config.hpp](https://github.com/kokkos/mdspan/blob/stable/include/experimental/__p0009_bits/config.hpp#L22-L27) and [macros.hpp](https://github.com/kokkos/mdspan/blob/stable/include/experimental/__p0009_bits/macros.hpp#L21). When compiler can find `version`, `type_traits` in macros.hpp can't be add if ignore system_header which already exist.

crtrott (approved):

PR #227, opened 2023-01-06
New submdspan CPP 17 compatible

crtrott:


crtrott:
This is the equivalent to #223 just with C++17 support

mhoemmen:
@youyu3 check this out

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -0,0 +1,332 @@
...
+  ExtentType extent;
+  StrideType stride;
+};
+
+namespace detail {
+
+// Mapping from submapping ranks to srcmapping ranks
+// InvMapRank is an index_sequence, which we build recursively
+// to contain the mapped indices.
+// MDSPAN_TEMPLATE_REQUIRES templates don't work for this
```
mhoemmen:
```suggestion
// MDSPAN_TEMPLATE_REQUIRES templates don't work for these
```

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -0,0 +1,199 @@
...
+    return mapping_offset<dst_mapping_t>{
+        dst_mapping_t(dst_ext),
+        static_cast<size_t>(src_mapping(detail::first_of(slices)...))};
+  } else {
+    // layout_stride case
+    #if MDSPAN_HAS_CXX_20
+    using inv_map_t = typename detail::inv_map_rank<0, index_sequence<>,
+                                                    SliceSpecifiers...>::type;
+    #else
+    using inv_map_t = decltype(detail::inv_map_rank(
```
mhoemmen: While the function version of `inv_map_rank` has a shorter and simpler implementation than the struct version, not everyone is a fan of functions whose only purpose is to deduce a type ("`decltype` everywhere").  Christian points out offline that we also don't know which approach would compile faster.  Thus, I think we should leave both of these paths in the code, at least until we have compilation benchmark results that suggest one over the other.

mhoemmen: Counterargument: We don't actually need `inv_map_t` here, we need an object of that type.  In that case, why not just call the function to get an `inv_map` object?

```c++
auto inv_map = detail::inv_map_rank(
  integral_constant<size_t,0>{},
  index_sequence<>{},
  (declval<SliceSpecifiers>())...);
return mapping_offset<dst_mapping_t>{
  dst_mapping_t(
    dst_ext,
    detail::construct_sub_strides(
      src_mapping,
      inv_map,
      tuple{detail::stride_of(slices)...})),
  static_cast<size_t>(src_mapping(detail::first_of(slices)...))};
```

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -0,0 +1,199 @@
...
+submdspan_mapping(const layout_right::mapping<Extents> &src_mapping,
+                  SliceSpecifiers... slices) {
+
+  // get sub extents
+  using src_ext_t = Extents;
+  auto dst_ext = submdspan_extents(src_mapping.extents(), slices...);
+  using dst_ext_t = decltype(dst_ext);
+
+  // determine new layout type
+  constexpr bool preserve_layout = detail::preserve_layout_right_mapping<
```
mhoemmen, line 208: It's a bit unfortunate that we can't write `preserve_layout_right_mapping` as a constexpr bool function (because some `SliceSpecifiers` may not be `constexpr`), given that we use it like one.

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -0,0 +1,199 @@
...
+        static_cast<size_t>(src_mapping(detail::first_of(slices)...))};
+  }
+}
+
+//**********************************
+// layout_right submdspan_mapping
+//*********************************
+namespace detail {
+
+// Figure out whether to preserve layout_left
```
mhoemmen:
```suggestion
// Figure out whether to preserve layout_right
```

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -0,0 +1,332 @@
...
+  size_t k, class Extents, class Slice,
+  /* requires */(is_convertible_v<Slice, tuple<size_t, size_t>>)
+)
+MDSPAN_INLINE_FUNCTION
+constexpr auto last_of(integral_constant<size_t, k>, const Extents &,
+                       const Slice &i) {
+  return get<1>(i);
+}
+
+// Suppress spurious warning with NVCC about no return statement.
```
mhoemmen, line 108: I've seen this with NVCC + GCC 10.  The GCC intrinsic function `__builtin_unreachable();` has helped me get rid of the warnings in this case.  C++23 adds `std::unreachable`, which hopefully will accomplish the same goal.

mhoemmen (approved):
EXCELLENT

Just one typo in a comment.

include/experimental/__p2630_bits/submdspan_extents.hpp
crtrott: Does that GCC intrinsic also work with nvc++, clang, msvc as host compiler?

mhoemmen: It's a GCC-only thing, so when I used it, I defined it as a macro.

mhoemmen: GCC 10 isn't quite as smart as GCC 11 about `constexpr auto` functions that have exhaustive `if constexpr` branches in them.

tests/test_submdspan.cpp
```diff
@@ -161,6 +194,62 @@ struct TestSubMDSpan<
...
+    int data[25000];
+    mds_org_t src(data, map);
+    size_t* result = allocate_array<size_t>(2);
+
+    dispatch([=] _MDSPAN_HOST_DEVICE () {
+      auto sub = stdex::submdspan(src, create_slice_arg(SubArgs())...);
+      bool match = match_expected_extents(0, 0, src.extents(), sub.extents(), create_slice_arg(SubArgs())...);
+      result[0] = match?1:0;
+    });
+    EXPECT_TRUE(result[0]==1);
```
dalg24: `
```suggestion
    EXPECT_EQ(result[0], 1);
```

tests/test_submdspan.cpp
```diff
@@ -161,6 +194,62 @@ struct TestSubMDSpan<
...
+    return (sub_ext.extent(sub_idx)==(p.extent+p.stride-1)/p.stride) && match_expected_extents(++src_idx, ++sub_idx, src_ext, sub_ext, slices...);
+  }
+  template<class SrcExtents, class SubExtents, class ... SliceArgs>
+  MDSPAN_INLINE_FUNCTION
+  static bool match_expected_extents(int src_idx, int sub_idx, SrcExtents src_ext, SubExtents sub_ext, stdex::full_extent_t, SliceArgs ... slices) {
+    return (sub_ext.extent(sub_idx)==src_ext.extent(src_idx)) && match_expected_extents(++src_idx, ++sub_idx, src_ext, sub_ext, slices...);
+  }
+  template<class SrcExtents, class SubExtents>
+  MDSPAN_INLINE_FUNCTION
+  static bool match_expected_extents(int, int, SrcExtents, SubExtents) { return true; }
```
dalg24: I am wondering whether we are we being to clever.  (I don't have a better idea.)

tests/test_submdspan.cpp
```diff
@@ -161,6 +194,62 @@ struct TestSubMDSpan<
...
+
+  static void run() {
+    typename mds_org_t::mapping_type map(typename mds_org_t::extents_type(ConstrArgs...));
+    int data[25000];
+    mds_org_t src(data, map);
+    size_t* result = allocate_array<size_t>(2);
```
dalg24:
```suggestion
    size_t* result = allocate_array<size_t>(1);
```

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -0,0 +1,200 @@
...
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#include <tuple>
+namespace std {
```
dalg24:
```suggestion
#include <array>
#include <type_traits>
#include <tuple>
#include <utility> // index_sequence

namespace std {
```

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -0,0 +1,315 @@
...
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
```
dalg24, line 16: include headers as appropriate

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -0,0 +1,315 @@
...
+template <class OffsetType, class ExtentType, class StrideType>
+struct strided_index_range {
+  using offset_type = OffsetType;
+  using extent_type = ExtentType;
+  using stride_type = StrideType;
+
+  OffsetType offset;
+  ExtentType extent;
+  StrideType stride;
+};
```
dalg24: Move it to its own header?

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -0,0 +1,315 @@
...
+}
+
+template <class OffsetType, class ExtentType, class StrideType>
+MDSPAN_INLINE_FUNCTION
+constexpr OffsetType
+first_of(const strided_index_range<OffsetType, ExtentType, StrideType> &r) {
+  return r.offset;
+}
+
+// last_of(slice): getting end of slice specifier range
```
dalg24, line 84: This comment is kinda misleading.  `last_of` takes more than one argument.  Why do `first_of` and `last_of` have different signatures by the way?  Of course I can read on and figure it out but the comment probably should answer that question.

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -0,0 +1,315 @@
...
+struct extents_constructor {
+  MDSPAN_TEMPLATE_REQUIRES(
+    class Slice, class... SliceSpecifiers,
+    /* requires */(!is_convertible_v<Slice, size_t> &&
+                   !is_strided_index_range<Slice>::value)
+  )
+  MDSPAN_INLINE_FUNCTION
+  constexpr static auto next_extent(const Extents &ext, const Slice &sl,
+                                    SliceSpecifiers... slices) {
+    constexpr size_t new_static_extent = StaticExtentFromRange<
```
mhoemmen, line 241: Would you consider making `StaticExtentFromRange` a `constexpr auto` function instead of a struct? That would make this code more concise, since you could get rid of the two `decltype` and the `::value` at the end.

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -0,0 +1,315 @@
...
+              SliceSpecifiers... slices) {
+    using index_t = typename Extents::index_type;
+    using new_static_extent_t =
+        StaticExtentFromStridedRange<ExtentType, StrideType>;
+    if constexpr (new_static_extent_t::value == dynamic_extent) {
+      using next_t =
+          extents_constructor<K - 1, Extents, NewExtents..., dynamic_extent>;
+      return next_t::next_extent(
+          ext, slices...,
+          r.extent > 0 ? 1 + divide<index_t>(r.extent - 1, r.stride) : 0);
```
mhoemmen, line 283: The code as it stands won't achieve the goal of `divide`, which is to return `integral_constant` if both inputs are `integral_constant`.

1. Adding 1 to `integral_constant<size_t, Value>` results in `size_t`, not `integral_constant<size_t, Value+1>`.
2. Even if adding 1 to `integral_constant` resulted in `integral_constant`, the result of the ternary operator would be `common_type_t<integral_constant<size_t, Value>, int>`, which is `int`.
  
https://godbolt.org/z/qWEdn5xM6

```c++
#include <utility>
#include <type_traits>

template<size_t Val1, size_t Val2>
constexpr auto divide(std::integral_constant<size_t, Val1> x,
  std::integral_constant<size_t, Val2> y) {
  return std::integral_constant<size_t, Val1 / Val2>{};
}

template<size_t Value>  
using Size = std::integral_constant<size_t, Value>;

int main() {
  constexpr auto z = divide(Size<8>{}, Size<2>{});
  static_assert(std::is_same_v<std::remove_const_t<decltype(z)>, Size<4>>);

  constexpr auto w = 1 + divide(Size<8>{}, Size<2>{});
  static_assert(not std::is_same_v<std::remove_const_t<decltype(w)>, Size<4>>);
  static_assert(std::is_same_v<std::remove_const_t<decltype(w)>, size_t>);

  static_assert(not std::is_same_v<std::common_type_t<Size<4>, int>, Size<4>>);
  static_assert(not std::is_same_v<std::common_type_t<Size<4>, int>, size_t>);
  static_assert(std::is_same_v<std::common_type_t<Size<4>, int>, int>);

  return 0;
}
```

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -0,0 +1,315 @@
...
+MDSPAN_INLINE_FUNCTION
+constexpr auto divide(const T0 &v0, const T1 &v1) {
+  return IndexT(v0) / IndexT(v1);
+}
+
+template <class IndexT, class T0, T0 v0, class T1, T1 v1>
+MDSPAN_INLINE_FUNCTION
+constexpr auto divide(const integral_constant<T0, v0> &,
+                      const integral_constant<T1, v1> &) {
+  return integral_constant<IndexT, v0 / v1>();
```
mhoemmen: Would you consider adding a `static_assert(v1 != 0);` before this line?  On the other hand, the compiler would produce an error in that case anyway, so perhaps we don't need to do anything here.

include/experimental/__p2630_bits/submdspan_extents.hpp
crtrott: This doesn't matter since it doesn't determine the type. Note the values I am generating here are used as constructor arguments for an extents type which is independently determined. In case static extents are generated or preserved, these values will at best be checked against the static_extents in the extents constructor.

include/experimental/__p2630_bits/submdspan_extents.hpp
crtrott: Tried it but if I really do this (i.e. just pass in slice etc. it turns out that the compiler things that that the function call is not a constant expression.

crtrott:
```
/home/crtrott/Kokkos/mdspan/include/experimental/__p2630_bits/submdspan_extents.hpp:236:22: error: 'sl' is not a constant expression
  236 |     constexpr size_t new_static_extent = static_extent_from_range(
```

crtrott: I guess I can just default construct the arguments ...

include/experimental/__p2630_bits/submdspan_extents.hpp
crtrott: done

include/experimental/__p2630_bits/submdspan_extents.hpp
crtrott: updated

include/experimental/__p2630_bits/submdspan_mapping.hpp
crtrott: done

include/experimental/__p2630_bits/submdspan_extents.hpp
crtrott: done

include/experimental/__p2630_bits/submdspan_extents.hpp
crtrott: I actually wrote a test for strided_index_range with zero extent and stride, and changed the logic here to support that.

include/experimental/__p2630_bits/submdspan_extents.hpp
crtrott: ah clang doesn't like this

crtrott: i.e. even though I just default construct pass the arguments in, which then get ignored. clang 12 didn't like this as constant expression.,

mhoemmen: @crtrott wrote:
> I guess I can just default construct the arguments ...

Just use `declval`, which you were doing already : - )

mhoemmen: Oh wait, I think `declval` won't work, because it's not an unevaluated context.  Oof....

include/experimental/__p2630_bits/submdspan_extents.hpp
mhoemmen: `integral_constant<T, Value>` implicitly converts to `T`.  This means that `1 + integral_constant<size_t, X>{} / integral_constant<size_t, Y>{}` just returns `1 + X / Y`, so you don't even need `divide` here.

https://godbolt.org/z/x48bnaW8e

include/experimental/__p2630_bits/submdspan_extents.hpp
mhoemmen: No worries then -- thanks!

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -0,0 +1,308 @@
...
+template <size_t K, class Extents, size_t... NewExtents>
+struct extents_constructor {
+  MDSPAN_TEMPLATE_REQUIRES(
+    class Slice, class... SliceSpecifiers,
+    /* requires */(!is_convertible_v<Slice, size_t> &&
+                   !is_strided_index_range<Slice>::value)
+  )
+  MDSPAN_INLINE_FUNCTION
+  constexpr static auto next_extent(const Extents &ext, const Slice &sl,
+                                    SliceSpecifiers... slices) {
```
nmm0: More of a code readability issue -- the slices and new extent parameters here are kind of mixed up into the same parameter pack so it's difficult to understand the code. Might be worth separating out the new extents into a separate parameter.

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -0,0 +1,308 @@
...
+  return IndexT(v0) / IndexT(v1);
+}
+
+template <class IndexT, class T0, T0 v0, class T1, T1 v1>
+MDSPAN_INLINE_FUNCTION
+constexpr auto divide(const integral_constant<T0, v0> &,
+                      const integral_constant<T1, v1> &) {
+  // cutting short division by zero
+  // this is used for strided_index_range with zero extent/stride
+  return integral_constant<IndexT, v0 == 0 ? 0 : v0 / v1>();
```
nmm0, line 189: I don't really understand this -- should stride be allowed to be zero? I guess if it is not specified it will be zero but then I think it would make sense to transform the value to nonzero before we get here.

nmm0 (approved):
Looks good to me,  just a few minor changes

crtrott:
I checked and this works now over my test suite of compilers/standards.

include/experimental/__p2630_bits/strided_index_range.hpp
```diff
@@ -0,0 +1,34 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+namespace std {
+namespace experimental {
+
+// Slice Specifier allowing for strides and compile time extent
+template <class OffsetType, class ExtentType, class StrideType>
+struct strided_index_range {
```
mhoemmen, line 23: This is called `strided_slice` now in the paper, but I'm OK with doing the rename as a separate PR.

include/experimental/__p2630_bits/strided_index_range.hpp
```diff
@@ -0,0 +1,34 @@
...
+  using offset_type = OffsetType;
+  using extent_type = ExtentType;
+  using stride_type = StrideType;
+
+  OffsetType offset;
+  ExtentType extent;
+  StrideType stride;
+};
```
mhoemmen, line 31: Should we consider enforcing the Mandates here?

```suggestion
  static_assert(is_integral_v<OffsetType> || is_integral_constant_v<OffsetType>);
  static_assert(is_integral_v<ExtentType> || is_integral_constant_v<ExtentType>);
  static_assert(is_integral_v<StrideType> || is_integral_constant_v<StrideType>);
};
```

In the change above, `is_integral_constant_v` is defined as follows.

```c++
template<class T>
struct is_integral_constant : std::false_type {};

template<std::integral T, T Value>
struct is_integral_constant<std::integral_constant<T, Value>> : std::true_type {};

template<class T>
constexpr inline bool is_integral_constant_v = is_integral_constant<T>::value;
```

include/experimental/__p2630_bits/submdspan_extents.hpp
mhoemmen: Thanks for figuring this out! : - )

mhoemmen (approved):
I think this is fine for now.  We can fix the `strided_slice` name and add Mandates checks later.  Thanks!

Issue #229, opened 2023-01-12
Line 132 of layout_right.hpp seems wrong


riclarsson:
Hi,

I think the line should simply read `decltype(other.stride(0)) stride = 1;`.  I guess you have that index type declared somewhere but I am not sure where.

The problem seems to only present itself when using the `full_extent_t` stuff with indices in a non-optimal way.  I need this because I am replacing code that has these features.  (I also need something like a partial_extent_t but I wrote my own for that for now because I could't find anything in the documentation.)

crtrott:
The most appropriate thing is probably a static cast to index_type (which is a member of the class). I also fixed the missing noexcept which is in the standard but was not in the implementation. The precondition check is now only done with NDEBUG not being defined.

crtrott:
@riclarsson can you check that this is fixed for you? Also what do you mean with `partial_extent_t`? What is the difference of that to just giving a `pair(begin,end)`?

riclarsson:
**About the updated code**

@crtrott I've since gone on to not allow turning a strided `mdspan` back into an exhaustive `mdspan` because it is simply better to not allow this if possible.  That said, I went back to my old code and your update does remove the warning about comparing different index sizes.  Thank you very much!

However, I think the solution is not optimal.  The following code-snippet demonstrates a new problem:

```cpp
  std::vector<double> v(6);
  for (std::int64_t i = 0; i < 6; i++)
    v[i] = static_cast<double>(1 + 10 * i);
  for (std::int64_t i = 0; i < 6; i++)
    std::cout << v[i] << ' ';
  std::cout << '\n';

  stdx::mdspan<double, stdx::dextents<std::int64_t, 2>> m(
      v.data(), std::array<std::int64_t, 2>{3, 2});
  for (std::int64_t i = 0; i < 3; i++)
    for (std::int64_t j = 0; j < 2; j++)
      std::cout << m(i, j) << ' ';
  std::cout << '\n';

  stdx::mdspan<double, stdx::dextents<std::int64_t, 1>, stdx::layout_stride>
      vs = stdx::submdspan(m, stdx::full_extent, 0);
  for (std::int64_t i = 0; i < 3; i++)
    std::cout << vs[i] << ' ';
  std::cout << '\n';

  stdx::mdspan<double, stdx::dextents<std::int64_t, 1>> ve(vs);
  for (std::int64_t i = 0; i < 3; i++)
    std::cout << ve[i] << ' ';
  std::cout << '\n';
```

[...]

PR #230, opened 2023-01-16
Address issue #229 and fix missing noexcept

crtrott:
The conversion operators from layout_stride for both layout_left and layout_right were missing a static_cast and also were not marked noexcept. I now guard the precondition check with NDEBUG.

Note: the throw in there is legal: it will result in a termination if triggered since the function is marked noexcept.

### DIFF
--- a/include/experimental/__p0009_bits/layout_left.hpp
+++ b/include/experimental/__p0009_bits/layout_left.hpp
@@ -116,18 +116,20 @@ class layout_left::mapping {
     )
     MDSPAN_CONDITIONAL_EXPLICIT((extents_type::rank() > 0))
     MDSPAN_INLINE_FUNCTION _MDSPAN_CONSTEXPR_14
-    mapping(layout_stride::mapping<OtherExtents> const& other) // NOLINT(google-explicit-constructor)
+    mapping(layout_stride::mapping<OtherExtents> const& other) noexcept // NOLINT(google-explicit-constructor)
       :__extents(other.extents())
     {
        /*
         * TODO: check precondition
         * other.required_span_size() is a representable value of type index_type
         */
-       #ifndef __CUDA_ARCH__
-       size_t stride = 1;
+       #if !defined(_MDSPAN_HAS_CUDA) && !defined(_MDSPAN_HAS_HIP) && !defined(NDEBUG)
+       index_type stride = 1;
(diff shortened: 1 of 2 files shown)


crtrott:
This addresses issue #229 hopefully.

include/experimental/__p0009_bits/layout_left.hpp
```diff
@@ -116,18 +116,19 @@ class layout_left::mapping {
...
         */
-       #ifndef __CUDA_ARCH__
-       size_t stride = 1;
+       #if !defined(_MDSPAN_HAS_CUDA) && !defined(_MDSPAN_HAS_HIP) && !defined(NDEBUG)
+       index_type stride = 1;
        for(rank_type r=0; r<__extents.rank(); r++) {
-         if(stride != other.stride(r))
+         if(stride != static_cast<index_type>(other.stride(r)))
            throw std::runtime_error("Assigning layout_stride to layout_left with invalid strides.");
+           // Note this throw will lead to a terminate if triggered since this function is marked noexcept
```
dalg24, line 130: why not just call `assert` in that case?

mhoemmen: To clarify @dalg24 's comment: `assert` does something exactly when `not defined(NDEBUG)` is `true`, so this should behave effectively like `assert`.  Furthermore, an actual `assert` may have a better error message, as a `noexcept` violation might not need to print the exception message.

The nicer way to do this would be to define an `MDSPAN_ASSERT` macro that Does The Right Thing, but I think it's fine for now just to use `assert`.

mhoemmen (approved):
I'm OK with merging this now, but I think @dalg24 has a good point that `assert` would better express intent.  It might also have a better error reporting experience.

include/experimental/__p0009_bits/layout_left.hpp
crtrott: I would need to write a function I call inside the assert I guess? Because I don't want to execute the loop unless !NDEBUG

Or do you just want me to replace the innermost if/throw?

dalg24: immediately invoked lambda code was worse.  reverted to the initial PR

dalg24 (approved):

PR #231, opened 2023-01-24
Update strided_index_range to strided_slice

crtrott:


include/experimental/__p2630_bits/strided_slice.hpp
```diff
@@ -15,19 +15,32 @@
...
+struct strided_slice {
   using offset_type = OffsetType;
   using extent_type = ExtentType;
   using stride_type = StrideType;
 
   OffsetType offset;
   ExtentType extent;
   StrideType stride;
+
+  static_assert(is_integral<OffsetType>::value || __mdspan_is_integral_constant<OffsetType>::value);
```
mhoemmen: Thanks for adding the Mandates! : - )

Optionally, given that the P2630 implementation only needs to support C++17, would you consider using `is_integral_v<T>` instead of `is_integral<T>::value`?

mhoemmen (approved):

include/experimental/__p2630_bits/strided_slice.hpp
crtrott: I think I got confused with the error message for is_integra_constant_v which doesn't exist :-)

crtrott: changed that back.

nmm0 (approved):
Looks pretty straightforward to me

PR #232, opened 2023-01-26
add CMakeUserPresets.txt to .gitignore

nmm0:


dalg24 (dismissed):
I would only accept `CMakeUserPresets.json` if accepting anything at all
as usual, this is something you can deal with locally on your system on on a particular repo.

mhoemmen (changes requested):
I'm with @dalg24 -- thanks for the contributions, though!

nmm0:
> I would only accept `CMakeUserPresets.json` if accepting anything at all as usual, this is something you can deal with locally on your system on on a particular repo.

Hmm, we do already have IDE specific folders such as that for visual studio and the generic build folder already on there. I can change to just CMakeUserPresets.json but do you want me to also remove the existing IDE-specific ignores?

mhoemmen:
> we do already have IDE specific folders such as that for visual studio and the generic build folder already on there

@dalg24 Those might just be for automated MSVC testing.  Do we have that somewhere?

crtrott (approved):

mhoemmen (approved):
I think this is fine.  Thanks for the contribution!

PR #233, opened 2023-01-27
This will make mdspan and its tests work with HIP

crtrott:
Note I didn't address benchmarks and examples, just tests. Also the biggest change is a patch gtest so it compiles with that clang version with C++20.


include/experimental/__p0009_bits/config.hpp
```diff
@@ -115,11 +115,15 @@ static_assert(_MDSPAN_CPLUSPLUS >= MDSPAN_CXX_STD_14, "mdspan requires C++14 or
 #  define _MDSPAN_NO_UNIQUE_ADDRESS
 #endif
 
+// AMDs HIP compiler seems to have issues with concepts
+// it pretends concepts exist, but doesn't ship <concept>
+#ifndef __HIPCC__
```
dalg24, line 120: But that would be defined regardless of whether it is amdclang++ or some other HIP-capable compiler.
Should we guard depending on whether HIP was enabled at configuration time? 
(not asking you to change, wondering what the right thing to do might be)
I see that we don't have such a "HIP enabled" macro define anyway.

include/experimental/__p0009_bits/macros.hpp
```diff
@@ -40,6 +40,16 @@
 #  define MDSPAN_INLINE_FUNCTION inline _MDSPAN_HOST_DEVICE
 #endif
 
+#ifndef MDSPAN_FUNCTION
+#  define MDSPAN_FUNCTION _MDSPAN_HOST_DEVICE
+#endif
+
+#ifdef _MDSPAN_HAS_HIP
+#  define MDSPAN_CTAD _MDSPAN_HOST_DEVICE
```
dalg24: I'd prefer CTAD -> DEDUCTION_GUIDE

include/experimental/__p0009_bits/mdspan.hpp
```diff
@@ -321,9 +321,16 @@ class mdspan
 
   MDSPAN_INLINE_FUNCTION
   friend constexpr void swap(mdspan& x, mdspan& y) noexcept {
+    // can't call the std::swap inside on HIP
+    #ifndef _MDSPAN_HAS_HIP
```
dalg24, line 325: is that a FIXME?  do you want to annotate with something greppable?

tests/test_mdspan_swap.cpp
```diff
@@ -102,7 +102,9 @@ TEST(TestMDSpanSwap, std_swap_dynamic_extents) {
   __MDSPAN_TESTS_RUN_TEST(test_mdspan_std_swap_dynamic_extents())
 }
 
-
+// On HIP we actually don't call through to swap via ADL
+// so the foo swap test which has sideeffects will fail
```
dalg24: sideeffects -> side effects

Consider annotating here too

tests/offload_utils.hpp
```diff
@@ -23,7 +30,23 @@ if (!(LHS == RHS)) { \
...
 
-#ifdef _MDSPAN_HAS_CUDA
+#if defined(_MDSPAN_HAS_CUDA) || defined(_MDSPAN_HAS_HIP)
+
+#if defined(_MDSPAN_HAS_CUDA)
+void deviceSynchronize() { (void) cudaDeviceSynchronize(); }
+template<class T>
+void mallocManaged(T** ptr, size_t size) { (void) cudaMallocManaged(ptr, size); }
+template<class T>
+void freeManaged(T* ptr) { (void) cudaFree(ptr); }
```
dalg24, line 40: I don't like the name but I suppose you were afraid of ambiguities?

include/experimental/__p0009_bits/config.hpp
crtrott: No for now whether HIPCC is defined is good enough. We can revisit later and refine based on compiler version when we have evidence which ones work.

include/experimental/__p0009_bits/macros.hpp
crtrott: sure

include/experimental/__p0009_bits/mdspan.hpp
crtrott: not really. There is nothing to fix, its just what we have to do.

tests/test_mdspan_swap.cpp
crtrott: as I said I don't think there is anything to fix?

tests/offload_utils.hpp
crtrott: partly, partly making clear that this is managed memory equivalent and that it needs to be that, because we use the allocations from both host and device in the tests.

include/experimental/__p0009_bits/mdspan.hpp
dalg24: do we consider it as a defect in the toolchain?

crtrott: Not really. I mean no one says we can call random std:: function on device ...

crtrott: That said it would be great if we could :-)

dalg24 (approved):
Please do a follow up PR where you pin the GTest version and apply the patch

PR #234, opened 2023-01-27
Enable testing with SYCL

masterleinad:
Corresponds to #223 but for SYCL.


tests/test_mdarray_ctors.cpp
```diff
@@ -181,9 +181,12 @@ void test_mdarray_ctor_data_carray() {
   free_array(errors);
 }
 
+// host data can't be used in device code
```
mhoemmen: What host data does the test use?  I'm guessing that the `mdarray` is defaulting to `vector` as its container; should we consider changing that to `array` so that the test can remain enabled, or does the test really depend on using `vector`?

masterleinad: Yes, I basically meant `std::vector`. I pushed a commit using `std::array` in the test case instead which works with `SYCL` but I would guess not with `Cuda` or `HIP`.

mhoemmen (approved):
Thanks for this change!

At some point, _after_ this PR merges : - ) , it would be excellent if we could refactor the testing infrastructure so that it could rely less on duplicated code protected by macros, and rely more on compile-time dispatch.

dalg24:
Will conflict with #233

crtrott:
SYCL testing doesn't seem to work inside the runner. But that wasn't really the intend of the runner anyway. So I think we should just revert the change to the test config thing.

crtrott (approved):

PR #235, opened 2023-01-30
Test with SYCL in the CI

masterleinad:
Based on top of #234. This pull request makes the `icpx` test with `SYCL`.
I also added `--output-on-failure` to `CTest` and needed to allow `CL/sycl.hpp` instead of `sycl/sycl.hpp` for older `oneAPI` releases. The header checks needs `C++17` but `SYCL` needs C++17 support anyway.


tests/offload_utils.hpp
```diff
@@ -15,7 +15,11 @@
 //@HEADER
 
 #ifdef _MDSPAN_HAS_SYCL
+#if __has_include(<sycl/sycl.hpp>)
 #include <sycl/sycl.hpp>
+#else
+#include <CL/sycl.hpp>
```
keryell: Perhaps this SYCL implementation does not provides this deprecated SYCL 1.2.1 header now?

masterleinad: Yeah, this one only had the deprecated header. Not quite sure if updating the compiler in the CI is worth it for this pull request.

nmm0 (approved):

Issue #200, opened 2022-10-20
Return type of `layout_stride::operator()` is incorrect


youyu3:
Should return `index_type` but returns `size_t`
https://github.com/kokkos/mdspan/blob/stable/include/experimental/__p0009_bits/layout_stride.hpp#L426

mhoemmen:
Confirmed, `layout_stride::mapping::operator()` should return `index_type` per the spec: http://eel.is/c++draft/views#mdspan.layout.stride.obs

crtrott:
@youyu3 can you issue this against this repo too?

youyu3:
@crtrott Sure I'll upstream my fixes to this repo when I get a chance.  Hopefully later this week.

PR #236, opened 2023-02-01
Fix return type of layout_stride operator, and add test

crtrott:
Addresses #200


dalg24 (approved):

compilation_tests/CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ add_compilation_test(ctest_constructor_sfinae)
 add_compilation_test(ctest_extents_ctors)
 add_compilation_test(ctest_extents_type_check)
 add_compilation_test(ctest_mdarray_type_check)
+add_compilation_test(ctest_layout_type_check)
```
mhoemmen, line 14: Yay for tests! : - D

compilation_tests/ctest_layout_type_check.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include "ctest_common.hpp"
+
+#include <experimental/mdspan>
+
+namespace stdex = std::experimental;
+
+using E1 = stdex::extents<int32_t, stdex::dynamic_extent, 3>;
+using MLR1 = stdex::layout_right::mapping<E1>;
+// Note we already separately tested that the extents interface is correct
+MDSPAN_STATIC_TEST(
```
mhoemmen, line 25: Optionally, would you consider making the test templated on the layout, so that you don't have to repeat all the conditions three times?  We'll have to add two more times due to P2642.

mhoemmen (approved):
EXCELLENT

compilation_tests/ctest_layout_type_check.cpp
crtrott: there is other places where we could do that, but I am not sure how to right now. Note that the macro only takes a single statement kinda.

Issue #201, opened 2022-10-20
Constraint on `layout_{left|right|stride}::stride` is not implemented


youyu3:
Per http://eel.is/c++draft/mdspan.layout.left.obs#5, should have
> `constexpr index_type stride(rank_type i) const;`
Constraints: `extents_­type​::​rank() > 0` is `true`

But is not implemented, for example: https://github.com/kokkos/mdspan/blob/stable/include/experimental/__p0009_bits/layout_left.hpp#L206

`stride(0)` returns 1 for a 0-D mapping (`stdex::extents<size_t>`).

youyu3:
One possible fix is to have the following
```
    MDSPAN_TEMPLATE_REQUIRES(
      class E = Extents,
      /* requires */ (
        E::rank() > 0
      )
    )
    MDSPAN_INLINE_FUNCTION
    constexpr index_type stride(rank_type i) const noexcept {
      ...
    }
```

youyu3:
`stride` isn't listed on http://eel.is/c++draft/mdspan.layout.stride.obs for `layout_stride` (it is on http://eel.is/c++draft/mdspan.layout.stride.overview), but I suppose it should have the same constraint?

mhoemmen:
@youyu3 We defined `layout_stride::mapping::stride` using the exposition-only `array<index_type, rank_> strides_`.  I suppose that one could be constrained too, but it doesn't need any constraints.  It gets its precondition -- that the input is in [0, `rank()`) -- indirectly from the `operator[]` [sequence container requirements](http://eel.is/c++draft/sequence.reqmts).

youyu3:
So is the behavior here https://godbolt.org/z/qWsnjeebT expected, i.e., `stride(0)` doesn't lead to a compilation error when the rank of the extent is 0?

mhoemmen:
That is not the correct behavior per http://eel.is/c++draft/views#mdspan.layout.left.obs .

crtrott:
@youyu3 can you issue that PR against kokkos/mdspan too?

PR #238, opened 2023-02-03
Add constraint for `layout_left|right|stride::stride()`, and add test

youyu3:
Addresses #201.
Had to update a few existing tests to get them to pass.

### DIFF
--- a/include/experimental/__p0009_bits/layout_left.hpp
+++ b/include/experimental/__p0009_bits/layout_left.hpp
@@ -176,14 +176,12 @@ class layout_left::mapping {
     MDSPAN_INLINE_FUNCTION constexpr bool is_exhaustive() const noexcept { return true; }
     MDSPAN_INLINE_FUNCTION constexpr bool is_strided() const noexcept { return true; }
 
-    MDSPAN_TEMPLATE_REQUIRES(
-      class E = Extents,
-      /* requires */ (
-        E::rank() > 0
-      )
-    )
     MDSPAN_INLINE_FUNCTION
-    constexpr index_type stride(rank_type i) const noexcept {
+    constexpr index_type stride(rank_type i) const noexcept
+#if MDSPAN_HAS_CXX_20
+      requires ( Extents::rank() > 0 )
+#endif
+    {
       index_type value = 1;
       for(rank_type r=0; r<i; r++) value*=__extents.extent(r);
       return value;
--- a/include/experimental/__p0009_bits/layout_left.hpp
+++ b/include/experimental/__p0009_bits/layout_left.hpp
@@ -176,6 +176,12 @@ class layout_left::mapping {
     MDSPAN_INLINE_FUNCTION constexpr bool is_exhaustive() const noexcept { return true; }
     MDSPAN_INLINE_FUNCTION constexpr bool is_strided() const noexcept { return true; }
 
+    MDSPAN_TEMPLATE_REQUIRES(
+      class E = Extents,
+      /* requires */ (
+        E::rank() > 0
+      )
+    )
     MDSPAN_INLINE_FUNCTION
     constexpr index_type stride(rank_type i) const noexcept {
       index_type value = 1;
--- a/tests/test_exhaustive_layouts.cpp
+++ b/tests/test_exhaustive_layouts.cpp
@@ -391,9 +391,11 @@ TYPED_TEST(TestLayoutConversion, implicit_conversion) {
   #endif
     map1 = typename TestFixture::map_1_t(map2);
 
-  for(size_t r=0; r != this->exts1.rank(); ++r) {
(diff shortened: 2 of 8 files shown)


tests/test_exhaustive_layouts.cpp
```diff
@@ -391,9 +391,11 @@ TYPED_TEST(TestLayoutConversion, implicit_conversion) {
   #endif
     map1 = typename TestFixture::map_1_t(map2);
 
-  for(size_t r=0; r != this->exts1.rank(); ++r) {
-    ASSERT_EQ(map1.extents().extent(r), map2.extents().extent(r));
-    ASSERT_EQ(map1.stride(r), map2.stride(r));
+  if constexpr (TestFixture::exts_1_t::rank() > 0) {
```
crtrott, line 395: We still test this with C++14

include/experimental/__p0009_bits/layout_left.hpp
```diff
@@ -176,6 +176,12 @@ class layout_left::mapping {
     MDSPAN_INLINE_FUNCTION constexpr bool is_exhaustive() const noexcept { return true; }
     MDSPAN_INLINE_FUNCTION constexpr bool is_strided() const noexcept { return true; }
 
+    MDSPAN_TEMPLATE_REQUIRES(
```
crtrott: I don't like introducing a template parameter here for C++17/14. I would say we just do this only for C++20 using a requires clause on a non-templated function, and note down in the readme that this constraint is not implemented pre-c++20,

tests/test_exhaustive_layouts.cpp
crtrott: Given my other comment, I would just protect the condition with whether C++20 is on.

tests/test_mdarray_ctors.cpp
```diff
@@ -143,15 +143,21 @@ void check_correctness(MDA& m, size_t rank, size_t rank_dynamic,
   ASSERT_EQ(m.rank_dynamic(), rank_dynamic);
   if(rank>0) {
     ASSERT_EQ(m.extent(0), extent_0);
-    ASSERT_EQ(m.stride(0), stride_0);
+    if constexpr (MDA::extents_type::rank()>0) {
```
crtrott: same as above.

crtrott (changes requested):
I don't really like introducing template parameters there. How about we just implement the constraint for C++20 and above, and then have the if constexpr condition and the specific test for availability protected also for C++20. Then note in the README that this constraint is not implemented in C++17/14.

tests/test_exhaustive_layouts.cpp
youyu3: Works for me.  Will update.

youyu3: Updated.  May not be the best-looking `#if`s...

mhoemmen: This code _should_ be fine even if `MDSPAN_HAS_CXX_20` defined, because the constraint on `stride(r)` won't come into play, no?  I'd rather have fewer macros if we can avoid it.

mhoemmen (approved):
I made a note in one place where I think an `#if ... #endif` could be removed.  Thanks!

tests/test_exhaustive_layouts.cpp
youyu3: This check is for the case when `exts_1_t::rank() == 0` in the C++20 mode, in which case we have the constraint and it'll result in a compilation error without the `if constexpr`.
Well, that being said, I think we should only protect the line that calls `stride()`.  I'll fix that.

tests/test_mdarray_ctors.cpp
```diff
@@ -143,15 +143,33 @@ void check_correctness(MDA& m, size_t rank, size_t rank_dynamic,
   ASSERT_EQ(m.rank_dynamic(), rank_dynamic);
   if(rank>0) {
     ASSERT_EQ(m.extent(0), extent_0);
-    ASSERT_EQ(m.stride(0), stride_0);
+#if MDSPAN_HAS_CXX_20
```
crtrott, line 144: can't we just move that above 144. After all runtime wise these checks anyway only execute if rank>0 and we don't need to check that it compiles, there are other tests which do taht.

crtrott: And I mean with that we only need one big constexpr if protection for all runtime check rank clauses.

crtrott (changes requested):
Just one small cleanup in the mdarray test, after that this is good.

tests/test_mdarray_ctors.cpp
youyu3: That makes sense.  Updated.

crtrott (approved):

PR #240, opened 2023-02-05
Mdarray to mdspan

crtrott:
Had to fix also the return type of the extents and mapping function (to what it already is in the latest paper revision),

### DIFF
--- a/include/experimental/__p1684_bits/mdarray.hpp
+++ b/include/experimental/__p1684_bits/mdarray.hpp
@@ -407,7 +407,7 @@ public:
   MDSPAN_INLINE_FUNCTION static constexpr rank_type rank_dynamic() noexcept { return extents_type::rank_dynamic(); }
   MDSPAN_INLINE_FUNCTION static constexpr size_t static_extent(size_t r) noexcept { return extents_type::static_extent(r); }
 
-  MDSPAN_INLINE_FUNCTION constexpr extents_type extents() const noexcept { return map_.extents(); };
+  MDSPAN_INLINE_FUNCTION constexpr const extents_type& extents() const noexcept { return map_.extents(); };
   MDSPAN_INLINE_FUNCTION constexpr index_type extent(size_t r) const noexcept { return map_.extents().extent(r); };
   MDSPAN_INLINE_FUNCTION constexpr index_type size() const noexcept {
 //    return __impl::__size(*this);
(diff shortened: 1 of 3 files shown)


include/experimental/__p1684_bits/mdarray.hpp
```diff
@@ -405,7 +407,7 @@ class mdarray {
   MDSPAN_INLINE_FUNCTION static constexpr rank_type rank_dynamic() noexcept { return extents_type::rank_dynamic(); }
   MDSPAN_INLINE_FUNCTION static constexpr size_t static_extent(size_t r) noexcept { return extents_type::static_extent(r); }
 
-  MDSPAN_INLINE_FUNCTION constexpr extents_type extents() const noexcept { return map_.extents(); };
+  MDSPAN_INLINE_FUNCTION constexpr const extents_type& extents() const noexcept { return map_.extents(); };
```
mhoemmen, line 410: This made me realize that [[mdspan.layout.reqmts]](http://eel.is/c++draft/mdspan.layout.reqmts) specifies that `.extents()` returns `const extents_type&`, but `layout_{left,right}_padded::mapping::extents` in P2642 returns `extents_type`.

I think that's OK here even for noncompliant layout mappings like `layout_{left,right}_padded::mapping`, because this is an example of lifetime extension, but we should talk about that in the context of P2642 at some point.

include/experimental/__p1684_bits/mdarray.hpp
```diff
@@ -72,6 +72,8 @@ class mdarray {
   using container_type = Container;
   using mapping_type = typename layout_type::template mapping<extents_type>;
   using element_type = ElementType;
```
mhoemmen, line 74: Would you consider implementing the Mandates on `ElementType` ("a complete object type that is neither an abstract class type nor an array type")?

```suggestion
  static_assert(std::is_same_v<std::remove_cv_t<ElementType>, ElementType> &&
    ! std::is_abstract_v<ElementType> &&
    ! std::is_array_v<ElementType>);
  using element_type = ElementType;
```

I think we can even check whether `ElementType` is a complete type, by tricks like [the one explained here](https://devblogs.microsoft.com/oldnewthing/20190710-00/?p=102678).

mhoemmen (approved):

include/experimental/__p1684_bits/mdarray.hpp
crtrott: For sure but I am trying to do a couple incremental updates. So will do that in another one.

PR #241, opened 2023-02-07
Bugfix: MSVC /Zc:preprocessor

MattEding:
Make MSVC compile when using a standards conforming preprocessor.
See https://learn.microsoft.com/en-us/cpp/preprocessor/preprocessor-experimental-overview


crtrott (approved):

mhoemmen (approved):
Thanks for the fix!

PR #242, opened 2023-02-11
Remove special gcc check

adah1972:
The current code is tested to work with major GCC versions (from GCC 7 on).

This will close issue #222.

### DIFF
--- a/include/experimental/__p0009_bits/config.hpp
+++ b/include/experimental/__p0009_bits/config.hpp
@@ -206,15 +206,6 @@ static_assert(_MDSPAN_CPLUSPLUS >= MDSPAN_CXX_STD_14, "mdspan requires C++14 or 
 #  endif
 #endif
 
-#ifndef _MDSPAN_USE_ALIAS_TEMPLATE_ARGUMENT_DEDUCTION
-// GCC 10's CTAD seems sufficiently broken to prevent its use.
-#  if (defined(_MDSPAN_COMPILER_CLANG) || !defined(__GNUC__) || __GNUC__ >= 11) \
-      && ((defined(__cpp_deduction_guides) && __cpp_deduction_guides >= 201907) \
-          || (!defined(__cpp_deduction_guides) && MDSPAN_HAS_CXX_20))
-#    define _MDSPAN_USE_ALIAS_TEMPLATE_ARGUMENT_DEDUCTION 1
-#  endif
-#endif
-
 #ifndef _MDSPAN_USE_STANDARD_TRAIT_ALIASES
 #  if (defined(__cpp_lib_transformation_trait_aliases) && __cpp_lib_transformation_trait_aliases >= 201304) \
           || (!defined(__cpp_lib_transformation_trait_aliases) && MDSPAN_HAS_CXX_14)
--- a/include/experimental/__p0009_bits/config.hpp
+++ b/include/experimental/__p0009_bits/config.hpp
@@ -198,10 +198,8 @@ static_assert(_MDSPAN_CPLUSPLUS >= MDSPAN_CXX_STD_14, "mdspan requires C++14 or 
 #endif
 
 #ifndef _MDSPAN_USE_CLASS_TEMPLATE_ARGUMENT_DEDUCTION
-// GCC 10's CTAD seems sufficiently broken to prevent its use.
-#  if (defined(_MDSPAN_COMPILER_CLANG) || !defined(__GNUC__) || __GNUC__ >= 11) \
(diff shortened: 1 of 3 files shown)


include/experimental/__p0009_bits/config.hpp
```diff
@@ -198,23 +198,12 @@ static_assert(_MDSPAN_CPLUSPLUS >= MDSPAN_CXX_STD_14, "mdspan requires C++14 or
 #endif
 
 #ifndef _MDSPAN_USE_CLASS_TEMPLATE_ARGUMENT_DEDUCTION
-// GCC 10's CTAD seems sufficiently broken to prevent its use.
-#  if (defined(_MDSPAN_COMPILER_CLANG) || !defined(__GNUC__) || __GNUC__ >= 11) \
-      && ((defined(__cpp_deduction_guides) && __cpp_deduction_guides >= 201703) \
-         || (!defined(__cpp_deduction_guides) && MDSPAN_HAS_CXX_17))
+#  if (defined(__cpp_deduction_guides) && __cpp_deduction_guides >= 201703) \
```
mhoemmen: @adah1972 Thank you for the contribution!  Have you tested this with GCC 8?  I've found that GCC 8 sometimes doesn't build C++17 code that works fine with GCC 7.5 or GCC 9.  The automated PR tests don't include a GCC 8 test.

adah1972: > @adah1972 Thank you for the contribution! Have you tested this with GCC 8? I've found that GCC 8 sometimes doesn't build C++17 code that works fine with GCC 7.5 or GCC 9. The automated PR tests don't include a GCC 8 test.

I am happy to report that all tests compile and pass with the current patch under GCC 8.5.0.

mhoemmen: Thanks for testing this!

mhoemmen (approved):
Thanks for the contribution and tests!

crtrott:
hold off this fails with NVCC

crtrott:
(we effectively implicitly disabled CTAD for most CUDA versions because most didn't support GCC11 anyway).

include/experimental/__p0009_bits/config.hpp
```diff
@@ -198,23 +198,13 @@ static_assert(_MDSPAN_CPLUSPLUS >= MDSPAN_CXX_STD_14, "mdspan requires C++14 or
 #endif
 
 #ifndef _MDSPAN_USE_CLASS_TEMPLATE_ARGUMENT_DEDUCTION
-// GCC 10's CTAD seems sufficiently broken to prevent its use.
-#  if (defined(_MDSPAN_COMPILER_CLANG) || !defined(__GNUC__) || __GNUC__ >= 11) \
-      && ((defined(__cpp_deduction_guides) && __cpp_deduction_guides >= 201703) \
-         || (!defined(__cpp_deduction_guides) && MDSPAN_HAS_CXX_17))
+#  if (!defined(__NVCC__) || (__CUDACC_VER_MAJOR__ >= 11 && __CUDACC_VER_MINOR__ >= 7)) && \
```
mhoemmen, line 201: Looks good! : - )

tests/test_layout_ctors.cpp
```diff
@@ -203,7 +203,8 @@ TEST(TestLayoutLeftListInitialization, test_layout_left_extent_initialization) {
   ASSERT_TRUE(m.is_exhaustive());
 }
 
-#if defined(_MDSPAN_USE_CLASS_TEMPLATE_ARGUMENT_DEDUCTION)
+// FIXME: CUDA NVCC including 12.0 does not like CTAD on nested classes
```
mhoemmen, line 206: It's not really an mdspan FIXME; it's an NVCC FIXME ; - )  Would you consider rewording?

```suggestion
// CUDA NVCC, including 12.0, implements CTAD but
// is not always able to perform CTAD on nested classes.
```

Another option would be to add a macro to config.hpp specifically for this case.  Then, you could comment the macro there once, so you wouldn't have to repeat the comment for each disabled test.

mhoemmen (approved):

include/experimental/__p0009_bits/config.hpp
mhoemmen: @youyu3 FYI

tests/test_layout_ctors.cpp
crtrott: The FIXME is just a reminder that we need to revisit this with later versions of CUDA.

crtrott (approved):

PR #246, opened 2023-03-13
Fix an ICE on MSVC 14.3X 

wmaxey:
These changes were ported from [libcudacxx.](https://github.com/NVIDIA/libcudacxx/pull/391)

MSVC has an issue parsing the aggregate expression comparing extents. Fixed by adding a small helper.

Left and right layout changes fixed a warning we were seeing in testing and seem to be conforming.

### DIFF
--- a/include/experimental/__p0009_bits/layout_left.hpp
+++ b/include/experimental/__p0009_bits/layout_left.hpp
@@ -163,7 +163,7 @@ class layout_left::mapping {
     )
     _MDSPAN_HOST_DEVICE
     constexpr index_type operator()(Indices... idxs) const noexcept {
-      return __compute_offset(__rank_count<0, extents_type::rank()>(), idxs...);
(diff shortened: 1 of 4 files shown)


include/experimental/__p0009_bits/layout_left.hpp
```diff
@@ -163,7 +163,7 @@ class layout_left::mapping {
     )
     _MDSPAN_HOST_DEVICE
     constexpr index_type operator()(Indices... idxs) const noexcept {
-      return __compute_offset(__rank_count<0, extents_type::rank()>(), idxs...);
+      return __compute_offset(__rank_count<0, extents_type::rank()>(), static_cast<index_type>(idxs)...);
```
mhoemmen, line 166: We discussed these changes.  They conform with the spec, as far as I can tell.

mhoemmen (approved):

mhoemmen (approved):
Thanks for the fix!

crtrott (approved):

Issue #202, opened 2022-10-25
Generic iteration function for n dimension mdspan 


SimeonEhrig:
Hi, I'm working on [vikunja](https://github.com/alpaka-group/vikunja), a platform-independent primitives library (e.g. `vikunja::transform`, `vikunja::reduce`) for different accelerators based on [alpaka](https://github.com/alpaka-group/alpaka). 
I want to use `mdspan` to support N dimension memory. In the presentation of Bryce Aldelstein Lelbach ([youtube link](https://www.youtube.com/watch?v=aFCLmQEkPUw&t=3124s)) there is the idea to build a recursive function to create n nested loops, which the compiler can optimize.

I implemented a prototype with `submdspan` for an iota function for a `mdspan`:   

```c++
namespace stdex = std::experimental;

template<int TDim>
struct Iterate_mdspan_impl;

template<>
struct Iterate_mdspan_impl<1>
{
    template<typename TSpan, typename TFunc>
    void operator()(TSpan span, TFunc& functor)
    {
        for(auto i = 0; i < span.extent(0); ++i)
        {
            span(i) = functor(span(i));
        }
    }
};

template<>
struct Iterate_mdspan_impl<2>
{
    template<typename TSpan, typename TFunc>
    void operator()(TSpan span, TFunc& functor)
    {
        for(auto i = 0; i < span.extent(0); ++i)
        {
            auto submdspan = stdex::submdspan(span, i, stdex::full_extent);
            Iterate_mdspan_impl<TSpan::rank() - 1>{}(submdspan, functor);
        }
    }
};

template<>
struct Iterate_mdspan_impl<3>
{
    template<typename TSpan, typename TFunc>
    void operator()(TSpan span, TFunc& functor)
    {
        for(auto i = 0; i < span.extent(0); ++i)
        {
            auto submdspan = stdex::submdspan(span, i, stdex::full_extent, stdex::full_extent);
            Iterate_mdspan_impl<TSpan::rank() - 1>{}(submdspan, functor);
        }
    }
};

template<typename TSpan, typename TData>
void iota_span(TSpan span, TData index)
{
    static_assert(TSpan::rank() > 0);
    static_assert(TSpan::rank() <= 3);
    auto functor = [&index](TData input) { return index++; };
    Iterate_mdspan_impl<TSpan::rank()>{}(span, functor);
}
```
[...]

PR #247, opened 2023-03-21
Add for_each_extents example

mhoemmen:
Add `for_each_extents` examples as two separate files.  This addresses common user questions about mdspan iteration, expressed in issues such as #202.

The first example relies on `std::views::cartesian_product` and `std::views::iota`, which are C++23 features that only GCC 13 currently implements.  The second example does not use ranges at all, but it does depend on lambdas being able to accept an explicit template parameter list, a C++20 feature.  The example could be rewritten not to depend on that, but it would be more verbose and generally less useful as an example in that case.  Both examples use macros to skip compilation in case the required features are not available.


examples/for_each_extents/for_each_extents.cpp
```diff
@@ -0,0 +1,115 @@
...
+  return concat_index_sequence(
+    reverse_index_sequence(std::index_sequence<Rest...>{}),
+    std::index_sequence<First>{});
+}
+
+template<size_t N>
+auto make_reverse_index_sequence()
+{
+  return reverse_index_sequence(std::make_index_sequence<N>());
+}
```
dalg24:
```suggestion
template <std::size_t... Is>
auto reverse(std::index_sequence<Is...>) -> std::index_sequence<sizeof...(Is) - 1 - Is...>;

template <std::size_t N>
using reverse_index_sequence = decltype(reverse(std::make_index_sequence<N>()));
```

mhoemmen: Much nicer -- thanks!

mhoemmen: Applied; thank you!

Issue #117, opened 2021-12-25
Are pitched allocations supported?


bernhardmgruber:
I spent some time playing with `mdspan` as abstraction to access multidimensional CPU/GPU buffers in [alpaka](https://github.com/alpaka-group/alpaka). You can think of these buffers as similar to `Kokkos::View`s. Now, higher-than-one dimensional buffers may employ padding between rows to align the start of each row at an address of a hardware specific multiple. This is analog to e.g. `cudaMallocPitch` in CUDA. I encountered several problems with `mdspan` in this regard:

The protocol between layout and access policy seems to deal in 1D element indices, not byte offsets, which are, however, needed when mapping an ND index into a 1D memory location with arbitrary gaps. The layout returns a 1D element index, the access policy accepts a 1D element index. P0009r14 seems a bit vage on that for the definition of layout mappings, but `m.required_span_size()`, which is used to define the valid range of values returned from `operator()`, specifies then "Returns: If Extents::rank() > 0 is true, the product of extents().extent(r) for all r in the range [0, Extents::rank() ). Otherwise, 1.", which to me sounds like you are not allowed to have empty space via unused indices in the span of memory underlying an `mdspan`. In the access policy however it reads very clearly that the 1D index is an element index and not a byte index: "access, a method to access the i-th element in the range of elements represented by a pointer;".

I could hack around that by definiting a custom layout policy which stored element size and row pitch and returned a 1D byte index instead of an element index. That in turn required a custom accessor that took a byte offset for the `access` and `offset` methods. That worked in principle, but I am sure that it is neither intended that way nor would it work with `submdspan`s/slicing. Furthermore, layout and access policy are now tighly coupled and you cannot combine them with other accessor or layout policies.
A different strategy could be to just define an accessor that divides the 1D index back into a row/col index and multiplies with the row pitch again. However, that would cost a division on each `mdspan` access. I have not tried this yet but it seems suboptimal to me as well.

This exploration raised a few questions:
* Does `mdspan` support pitched allocations at all? Maybe there is way which I have not found yet.
* Is the design of `mdspan` per P0009r14 incomplete and missing an important use case?
* If not: why does `mdspan` not support pitched allocations? They obviously exist, notable in CUDA, and I am certain that NVIDIA had a look at `mdspan`'s design. Pitched allocations also exist for CPUs (e.g. in-memory image or matrix formats). Is it because:
  - pitched allocations are a narrow and complicated use case and not worth the effort?
[...]

PR #249, opened 2023-03-21
Add pitched allocation example

### LINKED ISSUE #248: Add pitched allocation example
Add pitched allocation example from https://github.com/kokkos/mdspan/issues/117 .  The accessor in particular may be of interest.

mhoemmen:
This is the example I wrote for the user question here: https://github.com/kokkos/mdspan/issues/117

Fixes #248.


examples/pitched_allocation/pitched_allocation.cpp
```diff
@@ -0,0 +1,247 @@
...
+    //    
+    //return *reinterpret_cast<const T*>(p_);
+
+    T f;
+    std::memcpy(&f, p_, sizeof(T));
+    return f;
+  }
+
+private:
+  const char* p_ = nullptr;
```
dalg24: Declare it above to avoid going back with public and private access specifiers.

examples/pitched_allocation/pitched_allocation.cpp
```diff
@@ -0,0 +1,247 @@
...
+class const_proxy {
+private:
+  friend class const_tbs_proxy;
+  constexpr const_proxy(const char* p) noexcept
+    : p_(p)
+  {}
+
+public:
+  // Not constexpr because of reinterpret_cast or memcpy
+  operator T () const noexcept {
```
dalg24, line 71: If using C++20 you could do `bit_cast` and have the constexpr.

examples/pitched_allocation/pitched_allocation.cpp
```diff
@@ -0,0 +1,247 @@
...
+
+  std::uninitialized_fill_n(ptr, num_elements, tbs{1.0, 2, 3.0});
+
+  for(std::size_t k = 0; k < num_elements; ++k) {
+    assert(ptr[k].f0 == 1.0);
+    assert(ptr[k].i == 2);
+    assert(ptr[k].f1 == 3.0);
+  }
+
+  const tbs* ptr_c = ptr;
```
dalg24: Never used

dalg24 (commented):
Honestly I don't really follow the point you are trying to make with this example.

mhoemmen:
@dalg24 Thanks for reviewing! : - )

> Honestly I don't really follow the point you are trying to make with this example.

#117 explains in detail.  The point is that users sometimes need to fill a 2-D array where each contiguous segment (e.g., row for `layout_right`) has extra byte padding at the end.  Sometimes, the size of the type being stored does not evenly divide the number of bytes, even though the instances of the type are all stored with their correct alignment.

crtrott:
I do not understand how this does what you want with respect to pitched allocations?

crtrott:
I think the accessor needs to store a pitch or some information that every K elements there are 4 extra bytes or whatever.

crtrott:
Also your argument with the alignment doesn't make much sense to me. The object can't say that it is 12byte aligned anyway for pitched allocations, since the whole purpose is that for some rows its not.

crtrott:
Wouldn't this thing do what you want (only posted the relevant parts)?
```c++
template<class T>
struct pitched_accessor {
   int n_every;
   int extra_bytes;
   using data_handle_type = char*;
   pitched_accessor(int n_every_, int extra_bytes_) {
     assert(extra_bytes%alignof(T)==0);
   }
   T& access(const data_handle_type& p, size_t i) {
      char* p_offset = p + i*size_of(T) + extra_bytes * (i/n_every);
      return *reinterpret_cast<T&>(p_offset);
    }
 };
 ```
 Probably could have T* as `data_handle_type`, just reinterpret_cast to `char*` during access? Oh and offset can't be implemented with this accessor. It would need to store an extra thing to know how far into the first `n_every` you are offset.

examples/pitched_allocation/pitched_allocation.cpp
mhoemmen: Fixed; thanks!

examples/pitched_allocation/pitched_allocation.cpp
mhoemmen: Fixed; thanks!  I had to protect use of `bit_cast` with the appropriate feature test macro.  I did test locally with GCC 11.3, where the macro is enabled.

examples/pitched_allocation/pitched_allocation.cpp
mhoemmen: Fixed; thanks!

PR #251, opened 2023-04-04
add version file

nmm0:
This adds a VERSION file. Also, CMake will read the version from the file and it's easily read from scripts, so it will be easier to keep updated in releases

I therefore bumped the version to 0.5.0 since that's the last release (we forgot to update it from 0.4.0)


dalg24:
> This adds a VERSION file. Also, CMake will read the version from the file and it's easily read from scripts, so it will be easier to keep updated in releases

I don't get the argument this will be easier.

nmm0:
> > This adds a VERSION file. Also, CMake will read the version from the file and it's easily read from scripts, so it will be easier to keep updated in releases
> 
> I don't get the argument this will be easier.

You don't need to parse CMake to get the version from a tarball in a script.

I did some sleuthing and it's not super common to have a VERSION file, but pytorch does it (called version.txt) and CUDA has a version.json (which has a version for each component) in their tarball.

nmm0:
I'm closing this since there doesn't seem to be any desire for it.

PR #255, opened 2023-04-18
#254: qualify `std` namespace

### LINKED ISSUE #254: qualify `std::` namespace
Since we are currently embedding in namespace `std::` we make assumptions that standard library entities don't need to be qualified with `std::`. In order to move away from namespace `std::` we need to change that.

nmm0:
Closes #254

I honestly think I caught all of it. This builds if I change the mdspan namespace on gcc12 in C++14, 17, 20, and 23 modes. If I missed anything I'll catch it in the PR for changing the namespace...

### DIFF
--- a/include/experimental/__p0009_bits/default_accessor.hpp
+++ b/include/experimental/__p0009_bits/default_accessor.hpp
@@ -17,7 +17,7 @@
 
 #include "macros.hpp"
 
-#include <cstddef> // size_t
+#include <cstddef> // std::size_t
 
 namespace std {
 namespace experimental {
@@ -35,7 +35,7 @@ struct default_accessor {
   MDSPAN_TEMPLATE_REQUIRES(
     class OtherElementType,
     /* requires */ (
-      _MDSPAN_TRAIT(is_convertible, OtherElementType(*)[], element_type(*)[])
+      _MDSPAN_TRAIT(std::is_convertible, OtherElementType(*)[], element_type(*)[])
     )
   )
   MDSPAN_INLINE_FUNCTION
@@ -43,12 +43,12 @@ struct default_accessor {
 
   MDSPAN_INLINE_FUNCTION
   constexpr data_handle_type
-  offset(data_handle_type p, size_t i) const noexcept {
+  offset(data_handle_type p, std::size_t i) const noexcept {
     return p + i;
   }
 
   MDSPAN_FORCE_INLINE_FUNCTION
-  constexpr reference access(data_handle_type p, size_t i) const noexcept {
+  constexpr reference access(data_handle_type p, std::size_t i) const noexcept {
     return p[i];
   }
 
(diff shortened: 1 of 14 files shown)


include/experimental/__p0009_bits/mdspan.hpp
```diff
@@ -324,6 +324,7 @@ class mdspan
   friend constexpr void swap(mdspan& x, mdspan& y) noexcept {
     // can't call the std::swap inside on HIP
     #if !defined(_MDSPAN_HAS_HIP) && !defined(_MDSPAN_HAS_CUDA)
+    using std::swap;
```
nmm0, line 327: Note that this line was necessary for adl when the namespaces were renamed.

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -14,6 +14,8 @@
 //
 //@HEADER
 
+#include <tuple>
```
nmm0, line 17: To be honest, I don't know why this suddenly needed to be added but it's necessary or clang frontend compilers fail

include/experimental/__p0009_bits/default_accessor.hpp
```diff
@@ -35,20 +35,20 @@ struct default_accessor {
...
+      _MDSPAN_TRAIT(std::is_convertible, OtherElementType(*)[], element_type(*)[])
     )
   )
   MDSPAN_INLINE_FUNCTION
   constexpr default_accessor(default_accessor<OtherElementType>) noexcept {}
 
   MDSPAN_INLINE_FUNCTION
   constexpr data_handle_type
-  offset(data_handle_type p, size_t i) const noexcept {
+  offset(data_handle_type p, std::size_t i) const noexcept {
```
nliber: `size_t` is required to be in the global namespace due to C compatibility.

That being said, this doesn't hurt, so it is up to your discretion.

include/experimental/__p2630_bits/submdspan_extents.hpp
nliber: `tuple` is used.  Even if it was accidentally included via another system header, this is a better change.

nliber (approved):

include/experimental/__p0009_bits/default_accessor.hpp
nmm0: I see -- I kind of prefer it with the namespace but can remove it if people hate it.

crtrott: lets remove it again, makes for less change by a lot. just sed replace std::size_t with size_t :-)

crtrott (changes requested):
Yeah lets revert std::size_t to size_t, we generally also have been using int64_t etc. unqualified in Kokkos for example. Its just a lot of extra characters.

crtrott (approved):

PR #256, opened 2023-04-21
configurable namespaces

nmm0:
There are a ton of changes here, mostly because we use `std::experimental` extensively. This right now adds two macros, `MDSPAN_IMPL_STANDARD_NAMESPACE`, which defaults to `md` and `MDSPAN_IMPL_PROPOSED_NAMESPACE` which defaults to `experimental` and will be nested inside `MDSPAN_IMPL_STANDARD_NAMESPACE`.

These are of course, subject to change :)

I also added two headers for getting the "library"-style includes, `mdspan.hpp` and `experimental/mdarray.hpp`. I think eventually we probably want to remove the old `std`-style includes. I'm also considering moving these to a subfolder, e.g. `#include <mdspan/mdspan.hpp>` and `#include <mdspan/experimental/mdarray.hpp>`. Anyway, can adjust as needed.

Also, I avoided using CMake here so we can more easily copy the files over to Kokkos.


crtrott (changes requested):
That looks pretty good. One change: lets have mdspan and mdarray include mdspan.hpp/mdarray.hpp or the other way around so the likelyhood of errors is less. I.e. one of them simply defines the namespaces already.

include/experimental/__p0009_bits/layout_left.hpp
```diff
@@ -34,7 +33,7 @@ class layout_left::mapping {
     using layout_type = layout_left;
   private:
 
-    static_assert(detail::__is_extents_v<extents_type>, "std::experimental::layout_left::mapping must be instantiated with a specialization of std::experimental::extents.");
+    static_assert(detail::__is_extents_v<extents_type>, MDSPAN_PP_STRINGIFY(MDSPAN_IMPL_STANDARD_NAMESPACE) "::layout_left::mapping must be instantiated with a specialization of " MDSPAN_PP_STRINGIFY(MDSPAN_IMPL_STANDARD_NAMESPACE) "::extents.");
```
mhoemmen: Excellent : - )  (I'm glad that the error messages are parameterized, too!)

include/experimental/__p1684_bits/mdarray.hpp
```diff
@@ -20,8 +20,8 @@
 #include <cassert>
 #include <vector>
 
-namespace std {
-namespace experimental {
+namespace MDSPAN_IMPL_STANDARD_NAMESPACE {
+namespace MDSPAN_IMPL_PROPOSED_NAMESPACE {
```
mhoemmen, line 24: This forces mdarray to live in a nested namespace (e.g., `std::experimental`).  Users have no way to make it a single namespace without adding a `using` declaration or changing the code.

Should we consider adding a single macro `MDSPAN_IMPL_STANDARD_PROPOSED_NAMESPACE` that expresses the entire, possibly nested namespace?  It would default to `MDSPAN_IMPL_STANDARD_NAMESPACE::MDSPAN_IMPL_PROPOSED_NAMESPACE`.

On the other hand, if the namespaces aren't nested, it would complicate the implementation quite a bit.

mhoemmen (approved):
Thanks for doing this! : - )

include/experimental/__p1684_bits/mdarray.hpp
nmm0: Would it be enough to have some sort of toggle, like `MDSPAN_IMPL_PROPOSED_NAMESPACE_IS_NESTED`? The problem with a combined macro is that it won't work in C++14 mode.

mhoemmen: @nmm0 wrote:
> Would it be enough to have some sort of toggle, like `MDSPAN_IMPL_PROPOSED_NAMESPACE_IS_NESTED`?

The `mdarray` implementation currently assumes that it's living inside `MDSPAN_IMPL_STANDARD_NAMESPACE` and therefore doesn't need to namespace-qualify `mdspan` components like `extents`.  Fixing that, and testing the fix, would be a lot more effort.

> The problem with a combined macro is that it won't work in C++14 mode.

True, then you would need separate macros for "the name of the possibly nested namespace" (e.g., `std::experimental`), "open a scope in the existing namespace" (e.g., `namespace std { namespace experimental {`), and "close a scope in the existing namespace" (e.g., `}}`).

nmm0:
A couple of changes -- I put the main headers in `<mdspan/mdspan.hpp>` and `<mdspan/experimental/mdarray.hpp>`. I left all the other headers as is, but we can move them within `mdspan/` (except the std headers of course). I figured that would probably make the diff a bit difficult to track though.

Finally, we have the "std-like" mdspan header as `<experimental/mdspan>`, but mdspan is not in namespace `std` anymore. Do we want to change that and move it to be just `<mdspan>`? This would of course conflict with any actual standard library implementation and we would probably want a cmake option to disable that from happening in installs.

nmm0:
I also had to add an extra define `MDSPAN_IMPL_NAMESPACE_STD` when we are in "std-like" mode because of conflicts with `<span>`

crtrott:
Couple comments: I think I am ok to drop C++14 support, but I am also ok to NOT support nesting for things we consider mature. Also wouldn't it be a user thing, that if a user says I want everything in a nested namespace and they use C++17, they could define `MDSPAN_IMPL_STANDARD_NAMESPACE` as `std::experimental`, and then submdspan say would end up in `std::experimental::proposed`?

Also I have thought about defaults. I think `md` is too short and too potentially conflicted. I don't really see a reason why we shouldn't make the default "Kokkos" and "Experimental".

include/experimental/__p1684_bits/mdarray.hpp
crtrott: I am open to dropping support for 14 and then have MDSPAN_IMPL_

> I also had to add an extra define `MDSPAN_IMPL_NAMESPACE_STD` when we are in "std-like" mode because of conflicts with `<span>`

 Can't we fix that with a conditional include exits thingy?

crtrott: Or even just we don't define our own dynamic_extents when in C++20 mode ...

nmm0: I update to import `dynamic_extents` into the mdspan namespace.

nmm0:
> Couple comments: I think I am ok to drop C++14 support, but I am also ok to NOT support nesting for things we consider mature. Also wouldn't it be a user thing, that if a user says I want everything in a nested namespace and they use C++17, they could define `MDSPAN_IMPL_STANDARD_NAMESPACE` as `std::experimental`, and then submdspan say would end up in `std::experimental::proposed`?
> 
> Also I have thought about defaults. I think `md` is too short and too potentially conflicted. I don't really see a reason why we shouldn't make the default "Kokkos" and "Experimental".

@crtrott I changed the namespaces to Kokkos and Experimental

crtrott (approved):

PR #257, opened 2023-05-02
update README

nmm0:
This follows the changes in https://github.com/kokkos/mdspan/pull/256

FYI, github's "rich-diff" feature lets you view the diff of markdown files and how they are rendered (top right)


crtrott:
Add the cmake option.

PR #258, opened 2023-05-02
add option for stdmode install

nmm0:
Anyway, this PR adds a new CMake option, `MDSPAN_INSTALL_STDMODE_HEADERS` that will disable installing the std-like headers `<experimental/mdspan>` and `<experimental/mdarray>`. This will hopefully avoid conflicts with any standard library that actually ships mdspan if the user is not intending to use these headers (including ones that ship this library :P). I've defaulted this to `OFF` which is technically a breaking change so I can switch that around. However, in general I think it shouldn't be on by default.

Also, I realized that `<mdspan/mdspan.hpp>` and `<mdspan/mdarray.hpp>` were not installed, so this fixes that too.


dalg24 (approved):

crtrott:
You mean you defaulted to off right?

nmm0:
> You mean you defaulted to off right?

woops yeah

crtrott (approved):

crtrott:
I think this is ok.

PR #263, opened 2023-05-08
Fixup install mdspan include directory

### LINKED ISSUE #262: [CMake] install headers
After the [recent](https://github.com/kokkos/mdspan/pull/256) namespace updates, the cmake header install step [here](https://github.com/kokkos/mdspan/blob/stable/CMakeLists.txt#L168) seems to be missing installing the `include/mdspan` folder ?

dalg24:
Resolve #262 

The `include/mdspan/{mdarray,mdspan}.hpp` header files introduced in #256 were not installed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -165,7 +165,7 @@ export(TARGETS mdspan
     FILE mdspanTargets.cmake
 )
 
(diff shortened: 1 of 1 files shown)


mhoemmen (approved):
Thanks for the fix! : - )

Issue #273, opened 2023-06-07
mdspan fails to compile when compiler supports concepts but doesn't have <concepts>


J-Meyers:
Currently layout_stride.hpp has an include of \<concepts\> if `MDSPAN_HAS_CXX_20` and `MDSPAN_USE_CONCEPTS` is defined, `MD_SPAN_USE_CONCEPTS` only checks that the compiler supports concepts by checking `__cpp_concepts`, but does not check that the standard library includes concepts with `__cpp_lib_concepts`

Example of failure on godbolt: [https://godbolt.org/z/hejnYWxqq](https://godbolt.org/z/hejnYWxqq)

I wonder if this may be related to the comment on line 124 of config.hpp, godbolt does not seem to have the HIP compiler so I was unable to test

PR #274, opened 2023-06-07
Workaround for missing <concepts> but __cpp_concepts is defined

J-Meyers:
Fixes issue #273 

This approach to fixing the issue avoids the include of \<concepts\> entirely because the only thing needed from it was a very simple concept that could be recreated (same_as)

I put the new concept which is a duplicate of the standard one in the namespace `internal` rather than `detail` to avoid collisions with the standard library version which uses `detail` internally for its helper function `__same_as`

### DIFF
--- a/include/experimental/__p0009_bits/layout_stride.hpp
+++ b/include/experimental/__p0009_bits/layout_stride.hpp
@@ -30,8 +30,8 @@
 #ifdef __cpp_lib_span
 #include <span>
 #endif
-#if defined(_MDSPAN_USE_CONCEPTS) && MDSPAN_HAS_CXX_20
-#include <type_traits>  // for std::is_same_v
+#if defined(_MDSPAN_USE_CONCEPTS) && MDSPAN_HAS_CXX_20 && defined(__cpp_lib_concepts)
+#  include <concepts>
 #endif
 
 namespace MDSPAN_IMPL_STANDARD_NAMESPACE {
@@ -51,6 +51,7 @@ namespace detail {
     std::is_same<typename Layout::template mapping<typename Mapping::extents_type>, Mapping>::value;
 
 #if defined(_MDSPAN_USE_CONCEPTS) && MDSPAN_HAS_CXX_20
+#  if !defined(__cpp_lib_concepts)
   namespace internal {
   namespace detail {
   template <typename _Tp, typename _Up>
@@ -59,13 +60,20 @@ namespace detail {
   template <class T, class U>
   concept __same_as = detail::__same_as<T, U> && detail::__same_as<U, T>;
   } // namespace internal
+#  endif
 
   template<class M>
   concept __layout_mapping_alike = requires {
     requires __is_extents<typename M::extents_type>::value;
+#if defined(__cpp_lib_concepts)
+    { M::is_always_strided() } -> std::same_as<bool>;
+    { M::is_always_exhaustive() } -> std::same_as<bool>;
+    { M::is_always_unique() } -> std::same_as<bool>;
+#else
     { M::is_always_strided() } -> internal::__same_as<bool>;
     { M::is_always_exhaustive() } -> internal::__same_as<bool>;
     { M::is_always_unique() } -> internal::__same_as<bool>;
+#endif
     std::bool_constant<M::is_always_strided()>::value;
     std::bool_constant<M::is_always_exhaustive()>::value;
     std::bool_constant<M::is_always_unique()>::value;
--- a/include/experimental/__p0009_bits/layout_stride.hpp
+++ b/include/experimental/__p0009_bits/layout_stride.hpp
@@ -31,7 +31,7 @@
 #include <span>
 #endif
 #if defined(_MDSPAN_USE_CONCEPTS) && MDSPAN_HAS_CXX_20
-#include<concepts>
+#include <type_traits>  // for std::is_same_v
 #endif
 
 namespace MDSPAN_IMPL_STANDARD_NAMESPACE {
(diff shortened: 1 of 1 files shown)


mhoemmen:
@J-Meyers Thanks for the contribution!  I have actually encountered the situation you're describing, where a compiler (specifically MSVC, I think with Visual Studio 2019) knows the `concept` and `requires` syntax, but doesn't have the `<concepts>` header.

Some questions:

1. Does the compiler claim to implement C++20?  If so, then a reasonable objection to this PR is "we already support C++17, so we don't need a special case for compilers that claim to support C++20 but don't."

2. My experience with such versions of MSVC 2019 is that they don't have full support for the `requires` syntax.  It works in some places, but not others.  This means that supporting this use case is fragile.

3. The concepts _library_ paper [P0898R3](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0898r3.pdf) proposes the feature test macro ` __cpp_lib_concepts` for testing whether the concepts library is available.  (~There's unfortunately no feature test macro for "the compiler supports concepts and constraints syntax."~  CORRECTION: `__cpp_concepts` is that feature test macro; see https://en.cppreference.com/w/cpp/feature_test .)  Would you consider using this macro to decide whether the `<concepts>` header is available?  (If that doesn't work, then I would consider the compiler too broken to be worth supporting.)

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -31,7 +31,7 @@
 #include <span>
 #endif
 #if defined(_MDSPAN_USE_CONCEPTS) && MDSPAN_HAS_CXX_20
-#include<concepts>
+#include <type_traits>  // for std::is_same_v
 #endif
```
mhoemmen: `trait_backports.hpp` (included above) already includes `<type_traits>` unconditionally, so we don't need special logic to include it again here.  Also, would you consider only doing the work-around if the `<concepts>` header is not available, i.e., if the feature test macro `__cpp_lib_concepts` is not defined?

```c++
#if defined(_MDSPAN_USE_CONCEPTS) && MDSPAN_HAS_CXX_20 && defined(__cpp_lib_concepts)
#  include <concepts>
#endif
```

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -51,12 +51,21 @@ namespace detail {
...
 
 #if defined(_MDSPAN_USE_CONCEPTS) && MDSPAN_HAS_CXX_20
+  namespace internal {
+  namespace detail {
+  template <typename _Tp, typename _Up>
+  concept __same_as = std::is_same_v<_Tp, _Up>;
+  } // namespace detail
+  template <class T, class U>
+  concept __same_as = detail::__same_as<T, U> && detail::__same_as<U, T>;
+  } // namespace internal
```
mhoemmen: Would you consider only using the work-around if the `<concepts>` header is not available?  It's nicer to have the same symbol in the common case of "the compiler actually implements C++20 correctly."  Furthermore, the extra concepts may add to compile-time cost.

```suggestion
#if defined(_MDSPAN_USE_CONCEPTS) && MDSPAN_HAS_CXX_20
#  if ! defined(__cpp_lib_concepts)
  namespace internal {
  namespace detail {
  template <typename _Tp, typename _Up>
  concept __same_as = std::is_same_v<_Tp, _Up>;
  } // namespace detail
  template <class T, class U>
  concept __same_as = detail::__same_as<T, U> && detail::__same_as<U, T>;
  } // namespace internal
#  endif
```

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -51,12 +51,21 @@ namespace detail {
...
+
   template<class M>
   concept __layout_mapping_alike = requires {
     requires __is_extents<typename M::extents_type>::value;
-    { M::is_always_strided() } -> std::same_as<bool>;
-    { M::is_always_exhaustive() } -> std::same_as<bool>;
-    { M::is_always_unique() } -> std::same_as<bool>;
+    { M::is_always_strided() } -> internal::__same_as<bool>;
+    { M::is_always_exhaustive() } -> internal::__same_as<bool>;
+    { M::is_always_unique() } -> internal::__same_as<bool>;
```
mhoemmen, line 75:
```suggestion
#if defined(__cpp_lib_concepts)
    { M::is_always_strided() } -> std::same_as<bool>;
    { M::is_always_exhaustive() } -> std::same_as<bool>;
    { M::is_always_unique() } -> std::same_as<bool>;
#else
    { M::is_always_strided() } -> internal::__same_as<bool>;
    { M::is_always_exhaustive() } -> internal::__same_as<bool>;
    { M::is_always_unique() } -> internal::__same_as<bool>;
#endif
```

J-Meyers:
@mhoemmen 
> @J-Meyers Thanks for the contribution! I have actually encountered the situation you're describing, where a compiler (specifically MSVC, I think with Visual Studio 2019) knows the `concept` and `requires` syntax, but doesn't have the `<concepts>` header.
> 
> Some questions:
> 
> 1. Does the compiler claim to implement C++20?  If so, then a reasonable objection to this PR is "we already support C++17, so we don't need a special case for compilers that claim to support C++20 but don't."

The compiler's feature macros are accurate, it has the macro defined for `__cpp_concepts` and does support them, but the standard library does not define `__cpp_lib_concepts` and does not have the corresponding header \<concepts\>

> 2. My experience with such versions of MSVC 2019 is that they don't have full support for the `requires` syntax.  It works in some places, but not others.  This means that supporting this use case is fragile.
> 3. The concepts _library_ paper [P0898R3](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0898r3.pdf) proposes the feature test macro ` __cpp_lib_concepts` for testing whether the concepts library is available.  (~There's unfortunately no feature test macro for "the compiler supports concepts and constraints syntax."~  CORRECTION: `__cpp_concepts` is that feature test macro; see https://en.cppreference.com/w/cpp/feature_test .)  Would you consider using this macro to decide whether the `<concepts>` header is available?  (If that doesn't work, then I would consider the compiler too broken to be worth supporting.)

That makes sense, I committed what you suggested on 60f9138 I came to nearly the same design, I hadn't seen your comment [https://github.com/kokkos/mdspan/pull/274#discussion_r1223303873](https://github.com/kokkos/mdspan/pull/274#discussion_r1223303873) when I made my changes so the only difference is the spacing around the `! defined` where I didn't have the space between the `!` and the `defined` which I found consistent with the style in the rest of the project, but I can change it for emphasis if you think it would help.

include/experimental/__p0009_bits/layout_stride.hpp
mhoemmen: I normally recommend against this sort of macro-heavy C++ code, so please consider this and the change below optional.  I don't think "concept aliases" are a thing (by which I mean, "if `<concepts>` exists, alias `__same_as` to `std::same_as`, else define it like this").

mhoemmen (changes requested):
The only non-optional request I have is to consider using `__cpp_lib_concepts` to decide if the `<concepts>` header is available.  If that doesn't work with the compiler you have to support, then let's talk more : - ) .  Thanks!

mhoemmen:
I think our comments have overlapped in time a bit : - ) .  Thanks for the explanations!  I'll re-review.

mhoemmen (approved):
Thanks for the fix! : - )

mhoemmen:
> ... so the only difference is the spacing around the `! defined` where I didn't have the space between the `!` and the `defined` which I found consistent with the style in the rest of the project, but I can change it for emphasis if you think it would help.

No worries -- you did the right thing by being consistent with the project's style : - ) .  Thanks!!!

PR #279, opened 2023-07-12
fix extents converting constructor in constant expressions

oliverlee:


dalg24 (approved):

mhoemmen (approved):
Thanks for the fix! : - )

PR #281, opened 2023-08-12
Avoid reserved identifier collisions with newlib

### LINKED ISSUE #187: Build failing on GCC arm due to _N defined as a macro
`ctype.h` on that platform defines

```c
#define	_U	01
#define	_L	02
#define	_N	04
#define	_S	010
#define _P	020
#define _C	040
#define _X	0100
#define	_B	0200
```

which breaks `mdspan` in many places since some of those names are used as template parameters.

Names starting with an underscore and upper case letter are reserved by the standard to be used by an implementation so I think the proper way to handle this is rename the template parameters. I didn't find a quick way to automate the renaming otherwise I would have made a pull request.

ehntoo:
Newlib defines a macro _U in ctype.h (https://github.com/bminor/newlib/blob/048031501043e61ca31713b92ce2190213c7fb21/newlib/libc/include/ctype.h#L60-L67), resulting in a collision with the existing template names for __compressed_pair. Use _T1 and _T2 instead of _T and _U to avoid this collision. _T1 and _T2 are already used elsewhere, for example in type_list.hpp (https://github.com/kokkos/mdspan/blob/9d0a451e11177cbdeaef035c7914b0aa73ddd1e2/include/experimental/__p0009_bits/type_list.hpp#L76), so their use here should hopefully not introduce any new collisions with other environments.

Fixes #187.

Readme example compiler explorer demonstrating the original issue:
https://godbolt.org/z/s6qzafWP4

And a successful build made with this patch:
https://godbolt.org/z/zbMf4hddj

### DIFF
--- a/include/experimental/__p0009_bits/compressed_pair.hpp
+++ b/include/experimental/__p0009_bits/compressed_pair.hpp
@@ -27,16 +27,16 @@ namespace detail {
 
 // For no unique address emulation, this is the case taken when neither are empty.
 // For real `[[no_unique_address]]`, this case is always taken.
-template <class _T, class _U, class _Enable = void> struct __compressed_pair {
-  _MDSPAN_NO_UNIQUE_ADDRESS _T __t_val;
-  _MDSPAN_NO_UNIQUE_ADDRESS _U __u_val;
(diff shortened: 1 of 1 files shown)


mhoemmen (approved):
Thanks for the work-around! : - )

mhoemmen:
This should fix https://github.com/kokkos/mdspan/issues/187 , from what I understand.  Thanks for contributing! : - )

PR #282, opened 2023-10-16
Update workflow to clang++ instead of clang++12

crtrott:
Ubuntu test failed because clang++12 did not exist ..

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -16,7 +16,7 @@ jobs:
           compiler_prefix: /usr/bin
         - compiler_driver: g++-10
           compiler_prefix: /usr/bin
-        - compiler_driver: clang++-12
+        - compiler_driver: clang++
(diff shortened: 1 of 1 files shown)


.github/workflows/cmake.yml
```diff
@@ -16,7 +16,7 @@ jobs:
           compiler_prefix: /usr/bin
         - compiler_driver: g++-10
           compiler_prefix: /usr/bin
-        - compiler_driver: clang++-12
+        - compiler_driver: clang++
```
dalg24, line 19: What version does the latest ubuntu comes with?

dalg24: From the logs
```
-- The CXX compiler identification is Clang 14.0.0

crtrott: In my initial push I had changed it to `clang++-14` but that had failed too.

dalg24 (approved):

PR #284, opened 2023-10-17
Add tests from LLVM

crtrott:


crtrott:
Note I have not fixed yet discrepancies in our implementation. I disabled subtests with FIXME comments. There are 57 FIXME places.

crtrott:
Not sure if you want to check this in detail, I basically just copied the files over from LLVM, did a hacky test_macros.h implementation which imports stuff into std and defines some LLVM test macros, and disabled subtests which failed
I had to prefix a lot of the testfile names due to how our add test stuff is set up and the need to avoid duplicate target names

mhoemmen (approved):
These were ported straight from LLVM, per offline conversation with @crtrott today.

tests/libcxx-backports/extents/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+mdspan_add_test(comparison.pass)
+mdspan_add_test(conversion.pass)
+mdspan_add_test(ctad.pass)
+mdspan_add_test(ctor_default.pass)
+mdspan_add_test(ctor_from_array.pass)
+mdspan_add_test(ctor_from_integral.pass)
+#mdspan_add_test(ctor_from_span.pass)
```
dalg24, line 7: should that one be conditionally added when C++20 or later is used?

tests/libcxx-backports/extents/CtorTestCombinations.h
```diff
@@ -0,0 +1,99 @@
...
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+// UNSUPPORTED: c++03, c++11, c++14, c++17, c++20
+
+// <mdspan/mdspan.hpp>
+
+#include <mdspan/mdspan.hpp>
```
dalg24, line 12: is that the only change you did?

tests/libcxx-backports/extents/CtorTestCombinations.h
```diff
@@ -15,7 +15,7 @@
 #include <span>
 
 #include "../ConvertibleToIntegral.h"
-#include "test_macros.h"
+#include "../llvm_test_macros.h"
```
dalg24, line 18: Did you consider fiddling with the include paths instead of editing the header name?

tests/libcxx-backports/extents/CtorTestCombinations.h
```diff
@@ -45,15 +45,20 @@ constexpr void test_construction(AllExtents all_ext) {
 
   // test construction from just dynamic extents
   // create an array of just the extents corresponding to dynamic values
-  std::array<typename AllExtents::value_type, E::rank_dynamic()> dyn_ext{0};
-  size_t dynamic_idx = 0;
-  for (size_t r = 0; r < E::rank(); r++) {
-    if (E::static_extent(r) == std::dynamic_extent) {
-      dyn_ext[dynamic_idx] = all_ext[r];
-      dynamic_idx++;
+  if constexpr (E::rank_dynamic() > 0) {
```
dalg24, line 48: I don't understand.  What is going on here?

tests/CMakeLists.txt
```diff
@@ -68,5 +68,7 @@ mdspan_add_test(test_mdarray_to_mdspan)
 endif()
 
 if(CMAKE_CXX_STANDARD GREATER_EQUAL 20)
+if((CMAKE_CXX_COMPILER_ID STREQUAL Clang) OR ((CMAKE_CXX_COMPILER_ID STREQUAL GNU) AND (CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 12.0.0)))
```
dalg24, line 71: should it be `GREATER_EQUAL` -> `VERSION_GREATER_EQUAL`?

tests/libcxx-backports/MinimalElementType.h
```diff
@@ -25,20 +25,22 @@ struct MinimalElementType {
 template<class T, size_t N>
 struct ElementPool {
   constexpr ElementPool() {
-    ptr_ = std::allocator<T>().allocate(N);
+    nc_ptr_ = std::allocator<std::remove_const_t<T>>().allocate(N);
```
dalg24, line 28: I don't get this change either

tests/libcxx-backports/default_accessor/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+mdspan_add_test(default_accessor.access.pass)
+mdspan_add_test(default_accessor.ctor.conversion.pass)
+mdspan_add_test(default_accessor.ctor.default.pass)
+#mdspan_add_test(default_accessor.element_type.verify)
+mdspan_add_test(default_accessor.offset.pass)
+mdspan_add_test(default_accessor.types.pass)
```
dalg24, line 6: should we do a glob and detect the files ending in`. pass.cpp`?

tests/libcxx-backports/extents/CMakeLists.txt
crtrott: Yeah there is also another issue in there I have to track down. My plan is to do a cleanup PR where I actually fix issues etc.

tests/libcxx-backports/extents/CtorTestCombinations.h
crtrott: No see below

tests/libcxx-backports/extents/CtorTestCombinations.h
crtrott: Not really, but I guess we could.

tests/libcxx-backports/extents/CtorTestCombinations.h
crtrott: It didn't like the init and assignment to the array in the case of length zero. There are likely other ways to rewrite this if so desired.

tests/CMakeLists.txt
crtrott: ah probably VERSION_GREATER_EQUAL hm but its weird it did work on my system. I.e. I checked that with GNU 12 the tests were there and with 11 they were not.

tests/libcxx-backports/MinimalElementType.h
crtrott: libc++ accepts `std::allocator<const int>` libstdc++ says that is an invalid specialization of allocator.

tests/libcxx-backports/default_accessor/CMakeLists.txt
crtrott: possibly.

PR #286, opened 2023-10-27
Fix mdspan standard deviations

crtrott:


include/experimental/__p0009_bits/compressed_pair.hpp
```diff
@@ -28,8 +28,8 @@ namespace detail {
 // For no unique address emulation, this is the case taken when neither are empty.
 // For real `[[no_unique_address]]`, this case is always taken.
 template <class _T1, class _T2, class _Enable = void> struct __compressed_pair {
-  _MDSPAN_NO_UNIQUE_ADDRESS _T1 __t1_val;
-  _MDSPAN_NO_UNIQUE_ADDRESS _T2 __t2_val;
+  _MDSPAN_NO_UNIQUE_ADDRESS _T1 __t1_val{};
+  _MDSPAN_NO_UNIQUE_ADDRESS _T2 __t2_val{};
```
crtrott, line 32: Default initialization of nested members of mdspan, otherwise it would be not initialized in defaulted constructor.

include/experimental/__p0009_bits/compressed_pair.hpp
```diff
@@ -40,19 +40,19 @@ template <class _T1, class _T2, class _Enable = void> struct __compressed_pair {
   }
 
   MDSPAN_INLINE_FUNCTION_DEFAULTED
-  constexpr __compressed_pair() noexcept = default;
+  constexpr __compressed_pair() = default;
```
crtrott, line 43: noexcept ness for defaulted functions should be defaulted, required to have the expected behavior for the corresponding mdspan operations.

include/experimental/__p0009_bits/mdspan.hpp
```diff
@@ -55,6 +55,13 @@ class mdspan
     ReferenceType __callop(mdspan const& __self, const std::array<SizeType, N>& indices) noexcept {
       return __self.__accessor_ref().access(__self.__ptr_ref(), __self.__mapping_ref()(indices[Idxs]...));
     }
+#ifdef __cpp_lib_span
+    template <class ReferenceType, class SizeType, size_t N>
+    MDSPAN_FORCE_INLINE_FUNCTION static constexpr
+    ReferenceType __callop(mdspan const& __self, const std::span<SizeType, N>& indices) noexcept {
+      return __self.__accessor_ref().access(__self.__ptr_ref(), __self.__mapping_ref()(indices[Idxs]...));
+    }
+#endif
```
crtrott, line 64: This was just missing, even though the mdspan ctor called it. We just never tested it before ...

include/experimental/__p0009_bits/mdspan.hpp
```diff
@@ -183,6 +190,10 @@ class mdspan
       _MDSPAN_TRAIT(std::is_constructible, accessor_type, const OtherAccessor&)
     )
   )
+  MDSPAN_CONDITIONAL_EXPLICIT(
+    !_MDSPAN_TRAIT(std::is_convertible, const typename OtherLayoutPolicy::template mapping<OtherExtents>&, mapping_type) ||
+    !_MDSPAN_TRAIT(std::is_convertible, const OtherAccessor&, accessor_type)
+  )
```
crtrott, line 196: explicitness of the ctor was not specified while it should.

include/experimental/__p0009_bits/mdspan.hpp
```diff
@@ -226,8 +237,8 @@ class mdspan
   MDSPAN_TEMPLATE_REQUIRES(
     class SizeType,
     /* requires */ (
-      _MDSPAN_TRAIT(std::is_convertible, SizeType, index_type) &&
-      _MDSPAN_TRAIT(std::is_nothrow_constructible, index_type, SizeType)
+      _MDSPAN_TRAIT(std::is_convertible, const SizeType&, index_type) &&
+      _MDSPAN_TRAIT(std::is_nothrow_constructible, index_type, const SizeType&)
```
crtrott, line 241: wrong type used for constraints - the llvm tests check for the difference.

include/experimental/__p0009_bits/mdspan.hpp
```diff
@@ -311,7 +322,7 @@ class mdspan
   #endif // __cpp_lib_span
   #endif // MDSPAN_USE_PAREN_OPERATOR
 
-  MDSPAN_INLINE_FUNCTION constexpr size_t size() const noexcept {
+  MDSPAN_INLINE_FUNCTION constexpr size_type size() const noexcept {
```
crtrott, line 325: this returns size_type not size_t. Note that actually makes one of our old tests not work, I removed that thing.

include/experimental/__p0009_bits/mdspan.hpp
```diff
@@ -346,13 +357,13 @@ class mdspan
...
+  MDSPAN_INLINE_FUNCTION static constexpr bool is_always_exhaustive() { return mapping_type::is_always_exhaustive(); };
+  MDSPAN_INLINE_FUNCTION static constexpr bool is_always_strided() { return mapping_type::is_always_strided(); };
 
-  MDSPAN_INLINE_FUNCTION constexpr bool is_unique() const noexcept { return __mapping_ref().is_unique(); };
-  MDSPAN_INLINE_FUNCTION constexpr bool is_exhaustive() const noexcept { return __mapping_ref().is_exhaustive(); };
-  MDSPAN_INLINE_FUNCTION constexpr bool is_strided() const noexcept { return __mapping_ref().is_strided(); };
+  MDSPAN_INLINE_FUNCTION constexpr bool is_unique() const { return __mapping_ref().is_unique(); };
+  MDSPAN_INLINE_FUNCTION constexpr bool is_exhaustive() const { return __mapping_ref().is_exhaustive(); };
+  MDSPAN_INLINE_FUNCTION constexpr bool is_strided() const { return __mapping_ref().is_strided(); };
   MDSPAN_INLINE_FUNCTION constexpr index_type stride(size_t r) const { return __mapping_ref().stride(r); };
```
crtrott, line 367: These are not actually noexcept in the standard.

include/experimental/__p0009_bits/mdspan.hpp
```diff
@@ -374,7 +385,7 @@ class mdspan
 #if defined(_MDSPAN_USE_CLASS_TEMPLATE_ARGUMENT_DEDUCTION)
 MDSPAN_TEMPLATE_REQUIRES(
   class ElementType, class... SizeTypes,
-  /* requires */ _MDSPAN_FOLD_AND(_MDSPAN_TRAIT(std::is_integral, SizeTypes) /* && ... */) &&
+  /* requires */ _MDSPAN_FOLD_AND(_MDSPAN_TRAIT(std::is_convertible, SizeTypes, size_t) /* && ... */) &&
```
crtrott, line 388: wrong constraint on deduction guide.

tests/libcxx-backports/CustomTestLayouts.h
```diff
@@ -177,6 +177,7 @@ class layout_wrapping_integral<WrapArg>::mapping {
 
 #if MDSPAN_HAS_CXX_23
   friend constexpr void swap(mapping& x, mapping& y) noexcept {
+    using std::swap;
```
crtrott, line 180: extent is not in namespace std anymore, so need to make this findable.

tests/test_mdspan_size.cpp
```diff
@@ -58,9 +58,6 @@ TEST(TestMdspan, MdspanSizeReturnTypeAndPrecondition)
 
   static_assert(std::numeric_limits<std::int8_t>::max() == 127, "max int8_t != 127");
   test_mdspan_size(storage, Kokkos::extents<std::int8_t, 12, 11>{}); // 12 * 11 == 132
-
-  static_assert(std::numeric_limits<std::uint8_t>::max() == 255, "max uint8_t != 255");
-  test_mdspan_size(storage, Kokkos::extents<std::uint8_t, 16, 17>{}); // 16 * 17 == 272
```
crtrott, line 63: This isn't supposed to work according to standard, because size() returns size_type, the unsigned version of index_type

include/experimental/__p0009_bits/compressed_pair.hpp
```diff
@@ -28,8 +28,8 @@ namespace detail {
 // For no unique address emulation, this is the case taken when neither are empty.
 // For real `[[no_unique_address]]`, this case is always taken.
 template <class _T1, class _T2, class _Enable = void> struct __compressed_pair {
-  _MDSPAN_NO_UNIQUE_ADDRESS _T1 __t1_val;
-  _MDSPAN_NO_UNIQUE_ADDRESS _T2 __t2_val;
+  _MDSPAN_NO_UNIQUE_ADDRESS _T1 __t1_val{};
+  _MDSPAN_NO_UNIQUE_ADDRESS _T2 __t2_val{};
   MDSPAN_FORCE_INLINE_FUNCTION _MDSPAN_CONSTEXPR_14 _T1 &__first() noexcept { return __t1_val; }
```
mhoemmen, line 33: It looks like we still need `_MDSPAN_CONSTEXPR_14`, as `_MDSPAN_USE_CONSTEXPR_14` suggests the need for a work-around with Intel compilers.  Is that right?

https://github.com/search?q=repo%3Akokkos%2Fmdspan+_MDSPAN_USE_CONSTEXPR_14&type=code

include/experimental/__p0009_bits/compressed_pair.hpp
mhoemmen: Shouldn't `noexcept`-ness just be conditional on what the members do?

include/experimental/__p0009_bits/compressed_pair.hpp
mhoemmen: If neither type is empty in this specialization, why do you need `_MDSPAN_NO_UNIQUE_ADDRESS` at all?

mhoemmen (approved):

include/experimental/__p0009_bits/compressed_pair.hpp
crtrott: no idea I did NOT attempt to refactor anything here.

include/experimental/__p0009_bits/compressed_pair.hpp
crtrott: That is what implicitly happens for defaulted functions.

include/experimental/__p0009_bits/compressed_pair.hpp
crtrott: I think the comment above is the answer: if we do use real "unique address" attribute with compilers which support it, this is the only specialization of compressed pair which exists, i.e. we also use it for empty members.

include/experimental/__p0009_bits/compressed_pair.hpp
```diff
@@ -79,19 +79,19 @@ struct __compressed_pair<
   }
 
   MDSPAN_INLINE_FUNCTION_DEFAULTED
-  constexpr __compressed_pair() noexcept = default;
+  constexpr __compressed_pair() noexcept(noexcept(_T2())) = default;
```
dalg24: Is the noexcept expression actually needed here?

tests/libcxx-backports/mdspan/mdspan.index_operator.pass.cpp
```diff
@@ -38,6 +39,9 @@
 #include "../ConvertibleToIntegral.h"
 #include "../CustomTestLayouts.h"
 
+// This test uses the bracket operator, but its not something we have at configure time
+#if MDSPAN_USE_BRACKET_OPERATOR
```
nmm0, line 42: I think we should still test these with the operator() equivalents otherwise we may have holes in our test (e.g. C++20 span call but you mentioned above)

include/experimental/__p0009_bits/mdspan.hpp
nmm0: I'm confused -- wouldn't the be operator[] calling it? The mdspan.index_operator.pass.cpp test does call it, so I'm not entirely sure why it was being instantiated

EDIT: oh that test was disabled before this PR

nmm0 (commented):
Some initial comments

tests/libcxx-backports/mdspan/mdspan.index_operator.pass.cpp
```diff
@@ -38,6 +39,9 @@
...
+
 // Clang 16 does not support argument packs as input to operator []
 #if defined(__clang_major__) && __clang_major__ < 17
```
nmm0, line 45: Is this covered by `MDSPAN_USE_BRACKET_OPERATOR`?

nmm0 (dismissed):
I think there are a few instances where we are still using clang version macros to test for features such as https://github.com/kokkos/mdspan/pull/286/files#diff-1d6fc31052dd79dccd274528da7249a98a2ab178fc57087164a4752a64fc83b5R90-R94 that won't work with other compilers

tests/libcxx-backports/mdspan/mdspan.index_operator.pass.cpp
crtrott: no, and it was there before.

crtrott: more specifically, clang 16 and 15 can support most uses of bracket operator but not argument packs.

nmm0: Ok I misunderstood what that was there for

tests/libcxx-backports/mdspan/mdspan.index_operator.pass.cpp
crtrott: No I want to keep the libcxx-backports as close to the LLVM files as possible. I.e. the diffs should be minimal.

tests/libcxx-backports/mdspan/mdspan.swap.pass.cpp
```diff
@@ -43,11 +43,13 @@ constexpr void test_swap(MDS a, MDS b) {
   }
   // This check uses a side effect of layout_wrapping_integral::swap to make sure
   // mdspan calls the underlying components' swap via ADL
+  #if MDSPAN_HAS_CXX23
```
dalg24, line 46: Why 23 and not 20?

tests/libcxx-backports/mdspan/mdspan.index_operator.pass.cpp
```diff
@@ -23,6 +23,7 @@
 // Preconditions:
 //   * extents_type::index-cast(i) is a multidimensional index in extents_.
 
+
```
dalg24: Undo (unless that is a space that is present in the original file)

tests/test_mdspan_size.cpp
```diff
@@ -44,12 +44,12 @@ void test_mdspan_size(std::vector<char>& storage, Extents&& e)
   using extents_type = std::remove_cv_t<std::remove_reference_t<Extents>>;
   Kokkos::mdspan<char, extents_type> m(storage.data(), std::forward<Extents>(e));
 
-  static_assert(std::is_same<decltype(m.size()), std::size_t>::value,
-		"The return type of mdspan::size() must be size_t.");
+  static_assert(std::is_same<decltype(m.size()), typename decltype(m)::size_type>::value,
+		"The return type of mdspan::size() must be size_type.");
```
dalg24, line 49: Shouldn't we update the type of min_storage_size from size_t to size_type then?

tests/libcxx-backports/mdspan/mdspan.swap.pass.cpp
crtrott: because it needs consteval from C++23

dalg24: Oh right consteval if is C++23

tests/test_mdspan_size.cpp
crtrott: Sure

include/experimental/__p0009_bits/compressed_pair.hpp
crtrott: no

tests/libcxx-backports/mdspan/mdspan.index_operator.pass.cpp
crtrott: sure

tests/test_mdspan_size.cpp
dalg24: I don't understand that change/comment.
Everything else looks good.

crtrott: We previously returned size_t so mdspan::size() would work even if it overflowed uint8_t, but mdspan size return the unsigned version of index_type: so we can't return 272

dalg24: Wanna do 16, 15 then?

dalg24 (approved):

PR #302, opened 2023-11-27
Add element access via `at()` to `std::mdspan`

### LINKED ISSUE #300: Access with bounds checking using `::at`
In C++26, `std::span` will get a [`::at` member function](https://en.cppreference.com/w/cpp/container/span/at). It's probably a good idea to do this for mdspan as well. The function signatures would be the same as in [`operator[]`](https://en.cppreference.com/w/cpp/container/mdspan/operator_at):
```c++
template< class... OtherIndexTypes >
constexpr reference at( OtherIndexTypes... indices ) const;
```
```c++
template< class OtherIndexType >
constexpr reference at( std::span<OtherIndexType, rank()> indices ) const;
```
```c++
template< class OtherIndexType >
constexpr reference at( const std::array<OtherIndexType, rank()>& indices ) const;
```
Another neat advantage is that this does not require the multidimensional subscript operator, meaning that the signature is identical between in pre C++23 code.

Related discussion on std-proposals: https://lists.isocpp.org/std-proposals/2023/11/index.php#msg8301

stephanlachnit:
Closes #300

Possible implementation for ::at element access to mdspan with boundary checks. The boundary check is implemented with a for loop, as I did not see any other way to achieve this. The error message is inspired by the `std::span::at` error message from libstdc++ (https://github.com/gcc-mirror/gcc/commit/1fa85dcf656e2f2c7e483c9ed3c2680bf7db6858).

Note that this does not implement ::at element access to mdarray.

### DIFF
--- a/include/experimental/__p0009_bits/mdspan.hpp
+++ b/include/experimental/__p0009_bits/mdspan.hpp
@@ -24,6 +24,7 @@
 
 #include <stdexcept>
 #include <string>
+#include <type_traits>
 
 namespace MDSPAN_IMPL_STANDARD_NAMESPACE {
 template <
@@ -234,7 +235,7 @@ public:
   {
     size_t r = 0;
     for (const auto& index : {indices...}) {
-      if (index >= __mapping_ref().extents().extent(r)) {
+      if (__is_index_oor(index, __mapping_ref().extents().extent(r))) {
         throw std::out_of_range(
           "mdspan::at(...," + std::to_string(index) + ",...) out-of-range at rank index " + std::to_string(r) +
           " for mdspan with extent {...," + std::to_string(__mapping_ref().extents().extent(r)) + ",...}");
@@ -254,7 +255,7 @@ public:
   constexpr reference at(const std::array<SizeType, rank()>& indices) const
   {
     for (size_t r = 0; r < indices.size(); ++r) {
-      if (indices[r] >= __mapping_ref().extents().extent(r)) {
+      if (__is_index_oor(indices[r], __mapping_ref().extents().extent(r))) {
         throw std::out_of_range(
           "mdspan::at({...," + std::to_string(indices[r]) + ",...}) out-of-range at rank index " + std::to_string(r) +
           " for mdspan with extent {...," + std::to_string(__mapping_ref().extents().extent(r)) + ",...}");
@@ -274,7 +275,7 @@ public:
   constexpr reference at(std::span<SizeType, rank()> indices) const
   {
     for (size_t r = 0; r < indices.size(); ++r) {
-      if (indices[r] >= __mapping_ref().extents().extent(r)) {
+      if (__is_index_oor(indices[r], __mapping_ref().extents().extent(r))) {
         throw std::out_of_range(
           "mdspan::at({...," + std::to_string(indices[r]) + ",...}) out-of-range at rank index " + std::to_string(r) +
           " for mdspan with extent {...," + std::to_string(__mapping_ref().extents().extent(r)) + ",...}");
@@ -441,6 +442,23 @@ private:
   MDSPAN_FORCE_INLINE_FUNCTION constexpr mapping_type const& __mapping_ref() const noexcept { return __members.__second().__first(); }
   MDSPAN_FORCE_INLINE_FUNCTION _MDSPAN_CONSTEXPR_14 accessor_type& __accessor_ref() noexcept { return __members.__second().__second(); }
   MDSPAN_FORCE_INLINE_FUNCTION constexpr accessor_type const& __accessor_ref() const noexcept { return __members.__second().__second(); }
+  
+  MDSPAN_TEMPLATE_REQUIRES(
+    class SizeType,
+    /* requires */ (
+      _MDSPAN_TRAIT(std::is_convertible, const SizeType&, index_type) &&
+      _MDSPAN_TRAIT(std::is_nothrow_constructible, index_type, const SizeType&)
+    )
+  )
+  MDSPAN_FORCE_INLINE_FUNCTION constexpr bool __is_index_oor(SizeType index, index_type extent) const noexcept {
+    // Check for negative indices
+    if constexpr(std::is_signed_v<SizeType>) {
+      if(index < 0) {
+        return true;
+      }
+    }
+    return static_cast<index_type>(index) >= extent;
+  }
 
   template <class, class, class, class>
   friend class mdspan;
--- a/include/experimental/__p0009_bits/mdspan.hpp
+++ b/include/experimental/__p0009_bits/mdspan.hpp
@@ -18,6 +18,7 @@
 
 #include "default_accessor.hpp"
 #include "layout_right.hpp"
+#include "macros.hpp"
 #include "extents.hpp"
 #include "trait_backports.hpp"
 #include "compressed_pair.hpp"
@@ -442,7 +443,7 @@ private:
   MDSPAN_FORCE_INLINE_FUNCTION constexpr mapping_type const& __mapping_ref() const noexcept { return __members.__second().__first(); }
   MDSPAN_FORCE_INLINE_FUNCTION _MDSPAN_CONSTEXPR_14 accessor_type& __accessor_ref() noexcept { return __members.__second().__second(); }
   MDSPAN_FORCE_INLINE_FUNCTION constexpr accessor_type const& __accessor_ref() const noexcept { return __members.__second().__second(); }
-  
+
   MDSPAN_TEMPLATE_REQUIRES(
     class SizeType,
     /* requires */ (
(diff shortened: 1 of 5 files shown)


include/experimental/__p0009_bits/mdspan.hpp
```diff
@@ -218,6 +218,69 @@ class mdspan
...
 
+  MDSPAN_TEMPLATE_REQUIRES(
+    class... SizeTypes,
+    /* requires */ (
+      _MDSPAN_FOLD_AND(_MDSPAN_TRAIT(std::is_convertible, SizeTypes, index_type) /* && ... */) &&
+      _MDSPAN_FOLD_AND(_MDSPAN_TRAIT(std::is_nothrow_constructible, index_type, SizeTypes) /* && ... */) &&
+      (rank() == sizeof...(SizeTypes))
+    )
+  )
+  MDSPAN_FORCE_INLINE_FUNCTION
```
mhoemmen: Would you consider making this a host-only function?  Not all architectures that have `__host__ __device__` mark-up support throwing exceptions on device.

One could do this simply by omitting the `MDSPAN_FORCE_INLINE_FUNCTION` macro.

stephanlachnit: Done.

stigrs:
Given that SizeTypes can be signed, it should also check for indices[r] < 0.

stephanlachnit:
> Given that SizeTypes can be signed, it should also check for indices[r] < 0.

Thanks for this comment! Is this true for `std::span` as well? At least in `libstdc++`, they do not check this:

https://github.com/gcc-mirror/gcc/blob/fd7dabc116b9abc40ee6aa25bcc5d240b8cc516a/libstdc%2B%2B-v3/include/std/span#L290-L300

nmm0:
In `span`, I believe it's always just `std::size_t`, but `mdspan` has the size type as a template parameter for the extent and can be signed.

stephanlachnit:
> In `span`, I believe it's always just `std::size_t`, but `mdspan` has the size type as a template parameter for the extent and can be signed.

Ah right, thanks for pointing this out!

Edit: updated the pull request

include/experimental/__p0009_bits/mdspan.hpp
```diff
@@ -219,6 +222,68 @@ class mdspan
...
+      _MDSPAN_FOLD_AND(_MDSPAN_TRAIT(std::is_convertible, SizeTypes, index_type) /* && ... */) &&
+      _MDSPAN_FOLD_AND(_MDSPAN_TRAIT(std::is_nothrow_constructible, index_type, SizeTypes) /* && ... */) &&
+      (rank() == sizeof...(SizeTypes))
+    )
+  )
+  constexpr reference at(SizeTypes... indices) const
+  {
+    size_t r = 0;
+    for (const auto& index : {indices...}) {
+      if (index >= __mapping_ref().extents().extent(r) || index < 0) {
```
nmm0: Ugh sorry to complicate things further -- but I'm pretty sure some compilers will warn that this is a pointless comparison when index is size_t, so you (may) need to check for that. Also looks like you need to make sure the 0 is of the correct type (i.e. just cast to decltype(index)

stephanlachnit: Ah, you right, that makes sense. I opted for using a private function for this that check the out of range condition with a constexpr if to check `std::is_signed_v`.

include/experimental/__p0009_bits/mdspan.hpp
```diff
@@ -376,6 +442,23 @@ class mdspan
...
+      _MDSPAN_TRAIT(std::is_nothrow_constructible, index_type, const SizeType&)
+    )
+  )
+  MDSPAN_FORCE_INLINE_FUNCTION constexpr bool __is_index_oor(SizeType index, index_type extent) const noexcept {
+    // Check for negative indices
+    if constexpr(std::is_signed_v<SizeType>) {
+      if(index < 0) {
+        return true;
+      }
+    }
```
mhoemmen: `SizeType` doesn't need to be integral here or even less-than comparable with `int`.  mdspan behaves as if all operations on indices happen after conversion to `index_type`.  Thus, should we consider instead converting to `index_type` first, as we do below?

```suggestion
    // Check for negative indices
    if constexpr (std::is_signed_v<index_type>) {
      if (static_cast<index_type>(index) < 0) {
        return true;
      }
    }
```

stephanlachnit: Seems reasonable

stephanlachnit:
`if constexpr` doesn't exist pre C++17, should I add a macro here? Or just go with `if` and assume the compiler optimizes for this anyway?

include/experimental/__p0009_bits/mdspan.hpp
stephanlachnit: After thinking about it again, I'm not sure if this is correct: imagine I have `unsigned short` as `index_type` and an extent of size of `65535`. Now, I want to access with a `signed short` of value `-10`. What I end up getting is the element at index `65526`, which is not the expected result.

stephanlachnit: I ended up reverting this: if `SizeType` is not integral or less-than comparable with int, `std::is_signed_v` should be false anyway. The question is, what happens if `SizeType` can be negative, but is is not integral or less-than comparable with int, while `index_type` is not signed?

stephanlachnit:
I don't quite get the build failure on MSVC, would need some help there...

include/experimental/__p0009_bits/mdspan.hpp
mhoemmen: > After thinking about it again, I'm not sure if this is correct: imagine I have unsigned short as index_type and an extent of size of 65535. Now, I want to access with a signed short of value -10. What I end up getting is the element at index 65526, which is not the expected result.

Thanks for pointing this out!  The underlying issue, as I understand it, is that _`index-cast`_ returns its input (with the same type, not cast to `index_type`) if the input type is integral-not-bool.

[[mdspan.mdspan.members] 2 - 3](https://eel.is/c++draft/views.multidim#mdspan.mdspan.members-2) together say that the precondition on the input of `mdspan::operator[]` applies to `extents_type::`_`index-cast`_`(std::move(indices...))`.  For the specific case you mentioned of `SizeType = signed short`, [[mdspan.extents.expo] 9](https://eel.is/c++draft/views.multidim#mdspan.extents.expo-9) explains that _`index-cast`_ would just return `signed short`; it does not convert to `index_type`.  -10 cannot be part of a multidimensional index in `extents()`, so this violates a precondition.

mhoemmen: Btw, I was looking through [extents.hpp](https://github.com/kokkos/mdspan/blob/stable/include/experimental/__p0009_bits/extents.hpp) and noticed `check_all_indices`.  It looks like the layout mappings using this.  I think it could be refactored a bit to serve your needs.

include/experimental/__p0009_bits/mdspan.hpp
```diff
@@ -377,6 +444,23 @@ class mdspan
...
   MDSPAN_FORCE_INLINE_FUNCTION constexpr accessor_type const& __accessor_ref() const noexcept { return __members.__second().__second(); }
 
+  MDSPAN_TEMPLATE_REQUIRES(
+    class SizeType,
+    /* requires */ (
+      _MDSPAN_TRAIT(std::is_convertible, const SizeType&, index_type) &&
+      _MDSPAN_TRAIT(std::is_nothrow_constructible, index_type, const SizeType&)
+    )
+  )
+  MDSPAN_FORCE_INLINE_FUNCTION constexpr bool __is_index_oor(SizeType index, index_type extent) const noexcept {
```
nmm0, line 454: I'm of the opinion that we shouldn't add to the reserved names (i.e. `__` prefix) even though it's consistent with the rest of the class since we do plan on removing them.

nmm0:
> I don't quite get the build failure on MSVC, would need some help there...

What if you used the requires clause from `operator()`? I think that operator[] is not used in the MSVC build so we may not have updated it:

```cpp
  MDSPAN_TEMPLATE_REQUIRES(
    class... SizeTypes,
    /* requires */ (
      extents_type::rank() == sizeof...(SizeTypes) &&
      (detail::are_valid_indices<index_type, SizeTypes...>())
    )
  )
  ```

nmm0:
@stephanlachnit My apologies, we recently fixed a bunch of naming changes. Would you mind rebasing? If you don't have the time I can take this over just let me know. It would be very nice to have this merged in

stephanlachnit:
> @stephanlachnit My apologies, we recently fixed a bunch of naming changes. Would you mind rebasing? If you don't have the time I can take this over just let me know. It would be very nice to have this merged in

If you want to take over feel free to do so

nmm0:
Closing as it is superseded by https://github.com/kokkos/mdspan/pull/400

PR #306, opened 2024-01-04
Msvc layout stride fixes

crtrott:


include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -88,7 +88,7 @@ struct layout_stride {
     : private detail::__no_unique_address_emulation<
         detail::__compressed_pair<
           Extents,
-          std::array<typename Extents::index_type, Extents::rank()>
+          detail::possibly_empty_array<typename Extents::index_type, Extents::rank()>
```
crtrott, line 91: see below: the base_t typedef was already using possibly_empty_array. This needed to change, because std::array is not actually empty for rank == 0

include/experimental/__p0009_bits/config.hpp
```diff
@@ -231,7 +231,7 @@ static_assert(_MDSPAN_CPLUSPLUS >= MDSPAN_CXX_STD_14, "mdspan requires C++14 or
 #endif
 
 #ifndef MDSPAN_CONDITIONAL_EXPLICIT
-#  if MDSPAN_HAS_CXX_20 && !defined(_MDSPAN_COMPILER_MSVC)
```
crtrott, line 234: effectively this impacted some tests which didn't guard the convertibility tests correctly for C++20 & MSVC. But with current MSVC this just works anyway.

mhoemmen (approved):
EXCELLENT

PR #311, opened 2024-02-01
remove lambda expression in layout stride

oliverlee:
Replace a lambda expression to allow use in constant expression context in C++14.


mhoemmen:
Thanks for contributing this neat solution!

We'll have to decide whether we generally want to support C++14 any more, but I think this particular fix does not overly complicate the code.

Per discussion on https://github.com/kokkos/mdspan/issues/313 , the author may wish to add to this PR, so I won't merge it quite just yet.

mhoemmen (approved):

mhoemmen:
The author says it's OK to merge this PR.  Thanks for the contribution!

PR #314, opened 2024-02-10
Cpp14 support

oliverlee:
I've updated the P0009 headers to compile without C++17 extensions.

For the last commit, I've dumped a few common utilities in the extents header. Please feel free to edit as desired.


include/experimental/__p0009_bits/extents.hpp
```diff
@@ -16,11 +16,14 @@
...
 #pragma once
 #include "dynamic_extent.hpp"
+#include "fwd.hpp"
 
 #ifdef __cpp_lib_span
 #include <span>
 #endif
 #include <array>
+#include <type_traits>
+#include <stdexcept>
```
mhoemmen: Nothing in this header threw an `std::exception` subclass before.  Furthermore, the only reason this code now throws an exception is to trigger a call to `terminate` in a `noexcept` function.

Would you consider just calling `std::abort` as before?  Alternately, would you consider adding a configurable precondition violation handler (e.g., a macro that takes a message and just calls `std::abort` by default)?

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -367,6 +370,82 @@ struct maybe_static_array {
...
+template <class Layout, class Extents, class Mapping>
+constexpr void terminate_if_invalid_strides(with_rank<0>, Layout, const Extents&, const Mapping&)
+{}
+
+template <std::size_t N, class Layout, class Extents, class Mapping>
+constexpr void terminate_if_invalid_strides(with_rank<N>, Layout, const Extents& ext, const Mapping& other)
+{
+  static_assert(std::is_same<Layout, layout_left>::value or
+                std::is_same<Layout, layout_right>::value
+                , "");
```
mhoemmen: Would you consider adding a nonempty message here?

include/experimental/__p0009_bits/fwd.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+namespace MDSPAN_IMPL_STANDARD_NAMESPACE {
+namespace detail {
+
+// type used for rank-based tag dispatch
+//
+// this is used to enable alternatives to constexpr if when building for C++14
+//
+template <std::size_t N>
+struct with_rank {};
```
mhoemmen: Would you consider making this an alias of `std::integral_constant`, so as to avoid more instantiations of types?

mhoemmen:
Thanks for the contribution!  There are some nice refactorings in there as well.  @crtrott et al. will likely want to comment.

1. Do mdspan's tests and examples all build with C++14, or do these changes only just make your application build?  We're generally only interested in the former.

2. Would you consider adding C++14 testing to https://github.com/kokkos/mdspan/blob/stable/.github/workflows/cmake.yml as a part of this PR?  Current PR automatic testing will not exercise C++14.

include/experimental/__p0009_bits/extents.hpp
oliverlee: > Would you consider just calling std::abort as before? Alternately, would you consider adding a configurable precondition violation handler (e.g., a macro that takes a message and just calls std::abort by default)?

Sure that sounds like a better approach. I'm also wondering if `extents.hpp` is the right location for the `terminate_if_invalid_strides` function. Maybe `layout_stride.hpp` is a better place?

oliverlee:
>     1. Do mdspan's tests and examples all build with C++14, or do these changes only just make your application build?  We're generally only interested in the former.

I've only tested with the tests defined in CMake along with C++14 so far. All tests pass before and after this PR - however after this PR now builds without warnings.

>     2. Would you consider adding C++14 testing to https://github.com/kokkos/mdspan/blob/stable/.github/workflows/cmake.yml as a part of this PR?  Current PR automatic testing will not exercise C++14.

Sure.

mhoemmen:
Thanks for making the changes!  Building with C++14 results in the following CMake error message.

```
CMake Error at CMakeLists.txt:200 (MESSAGE):
  Benchmarks are not available in C++14 or C++23 mode.  Turn
  MDSPAN_ENABLE_BENCHMARKS OFF or use C++17 or C++20.
```

I think you could just change line 200 of CMakeLists.txt ( https://github.com/kokkos/mdspan/blob/0e6a69dfe045acbb623003588a4aff844ea4b276/CMakeLists.txt#L200 ) from what it is currently,

```
MESSAGE(FATAL_ERROR "Benchmarks are not available in C++14 or C++23 mode. Turn MDSPAN_ENABLE_BENCHMARKS OFF or use C++17 or C++20.")
```

to something like

```
set(MDSPAN_ENABLE_BENCHMARKS OFF)
MESSAGE(STATUS "Benchmarks are not available in C++14 or C++23 mode. Turn MDSPAN_ENABLE_BENCHMARKS OFF or use C++17 or C++20.")
```

oliverlee:
> Thanks for making the changes! Building with C++14 results in the following CMake error message.
> 
> ```
> CMake Error at CMakeLists.txt:200 (MESSAGE):
>   Benchmarks are not available in C++14 or C++23 mode.  Turn
>   MDSPAN_ENABLE_BENCHMARKS OFF or use C++17 or C++20.
> ```
> 
> I think you could just change line 200 of CMakeLists.txt (
> 
> https://github.com/kokkos/mdspan/blob/0e6a69dfe045acbb623003588a4aff844ea4b276/CMakeLists.txt#L200
> ) from what it is currently,
> 
> ```
> MESSAGE(FATAL_ERROR "Benchmarks are not available in C++14 or C++23 mode. Turn MDSPAN_ENABLE_BENCHMARKS OFF or use C++17 or C++20.")
> ```
> 
> to something like
> 
> ```
> set(MDSPAN_ENABLE_BENCHMARKS OFF)
> MESSAGE(STATUS "Benchmarks are not available in C++14 or C++23 mode. Turn MDSPAN_ENABLE_BENCHMARKS OFF or use C++17 or C++20.")
> ```

I disabled benchmarking for C++14 using the same approach for C++23 (i.e. via matrix settings in `cmake.yml`) since I thought it would be simpler to maintain. Let me know if you still prefer a CMake option based approach.

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -537,14 +539,9 @@ template <class IndexType, size_t... Extents> class extents {
...
-    if constexpr (rank() != extents<OtherIndexType, OtherExtents...>::rank()) {
-      return false;
-    } else {
-      using common_t = std::common_type_t<index_type, OtherIndexType>;
-      for (size_type r = 0; r < m_rank; r++)
-        if(static_cast<common_t>(rhs.extent(r)) != static_cast<common_t>(lhs.extent(r))) return false;
-    }
-    return true;
+    return
+      rank() == extents<OtherIndexType, OtherExtents...>::rank() and
```
crtrott:
```suggestion
      rank() == extents<OtherIndexType, OtherExtents...>::rank() &&
```

include/experimental/__p0009_bits/macros.hpp
```diff
@@ -101,6 +103,52 @@
...
+{
+  (void)std::fprintf(::stderr, "%s:%u: precondition failure: `%s`\n", file, line, cond);
+  std::abort();
+}
+
+} // namespace detail
+} // namespace MDSPAN_IMPL_STANDARD_NAMESPACE
+
+#ifndef MDSPAN_PRECONDITION_VIOLATION_HANDLER
+#define MDSPAN_PRECONDITION_VIOLATION_HANDLER(cond, file, line) \
```
crtrott:
```suggestion
#ifndef MDSPAN_IMPL_PRECONDITION_VIOLATION_HANDLER
#define MDSPAN_IMPL_PRECONDITION_VIOLATION_HANDLER(cond, file, line) \
```

include/experimental/__p0009_bits/macros.hpp
```diff
@@ -101,6 +103,52 @@
...
+  MDSPAN_IMPL_STANDARD_NAMESPACE::detail::default_precondition_violation_handler(cond, file, line)
+#endif
+
+#ifndef MDSPAN_SKIP_PRECONDITION_VIOLATION_HANDLER
+  #ifndef NDEBUG
+    #define MDSPAN_SKIP_PRECONDITION_VIOLATION_HANDLER 0
+  #else
+    #define MDSPAN_SKIP_PRECONDITION_VIOLATION_HANDLER 1
+  #endif
+#endif
```
crtrott:
```suggestion
#ifndef MDSPAN_IMPL_SKIP_PRECONDITION_VIOLATION_HANDLER
  #ifndef NDEBUG
    #define MDSPAN_IMPL_SKIP_PRECONDITION_VIOLATION_HANDLER 0
  #else
    #define MDSPAN_IMPL_SKIP_PRECONDITION_VIOLATION_HANDLER 1
  #endif
#endif
```

include/experimental/__p0009_bits/macros.hpp
```diff
@@ -101,6 +103,52 @@
...
+
+namespace MDSPAN_IMPL_STANDARD_NAMESPACE {
+namespace detail {
+
+template <bool skip = MDSPAN_SKIP_PRECONDITION_VIOLATION_HANDLER>
```
crtrott:
```suggestion
template <bool skip = MDSPAN_IMPL_SKIP_PRECONDITION_VIOLATION_HANDLER>
```

include/experimental/__p0009_bits/macros.hpp
```diff
@@ -101,6 +103,52 @@
...
+{
+  if (skip) { return; }
+
+  MDSPAN_PRECONDITION_VIOLATION_HANDLER(cond, file, line);
+}
+
+} // namespace detail
+} // namespace MDSPAN_IMPL_STANDARD_NAMESPACE
+
+#define MDSPAN_PRECONDITION(...) \
```
crtrott:
```suggestion
#define MDSPAN_IMPL_PRECONDITION(...) \
```

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -560,4 +574,34 @@ struct layout_stride {
...
+
+  constexpr auto is_left = std::is_same<Layout, layout_left>::value;
+
+  typename Extents::index_type stride = 1;
+
+  for (std::size_t r = 0; r < N; r++) {
+    const std::size_t s = is_left ? r : N - 1 - r;
+
+    MDSPAN_PRECONDITION(common_integral_compare(stride, other.stride(s))
+                        and "invalid strides for layout_{left,right}");
```
crtrott:
```suggestion
    MDSPAN_IMPL_PRECONDITION(common_integral_compare(stride, other.stride(s))
                             and "invalid strides for layout_{left,right}");
```

tests/test_alternate_precondition_violation_handler.cpp
```diff
@@ -0,0 +1,29 @@
+#include <stdexcept>
+
+#define MDSPAN_PRECONDITION_VIOLATION_HANDLER(cond, file, line) \
+  do { \
+    throw std::logic_error{"precondition failure"}; \
+  } while (0);
```
crtrott:
```suggestion
#define MDSPAN_IMPL_PRECONDITION_VIOLATION_HANDLER(cond, file, line) \
  do { \
    throw std::logic_error{"precondition failure"}; \
  } while (0);
```

tests/test_alternate_precondition_violation_handler.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+#include <mdspan/mdspan.hpp>
+#include <gtest/gtest.h>
+
+TEST(mdspan_macros, alternate_precondition_violation_handler)
+{
+  ASSERT_THROW(MDSPAN_PRECONDITION(false), std::logic_error);
```
crtrott:
```suggestion
  ASSERT_THROW(MDSPAN_IMPL_PRECONDITION(false), std::logic_error);
```

tests/test_alternate_precondition_violation_handler.cpp
```diff
@@ -0,0 +1,29 @@
...
+}
+
+TEST(mdspan_macros, alternate_precondition_check_constexpr_invocable)
+{
+  struct fn
+  {
+    constexpr auto operator()() const
+    {
+      MDSPAN_PRECONDITION(1 + 1 == 2);
```
crtrott:
```suggestion
      MDSPAN_IMPL_PRECONDITION(1 + 1 == 2);
```

tests/test_macros.cpp
```diff
@@ -0,0 +1,24 @@
+#include <mdspan/mdspan.hpp>
+#include <gtest/gtest.h>
+
+TEST(mdspan_macros, precondition_violation)
+{
+  constexpr auto msg = "hello, world!";
+
+  ASSERT_DEATH(MDSPAN_PRECONDITION(false and "hello, world!"), msg);
```
crtrott:
```suggestion
  ASSERT_DEATH(MDSPAN_IMPL_PRECONDITION(false and "hello, world!"), msg);
```

tests/test_macros.cpp
```diff
@@ -0,0 +1,24 @@
...
+}
+
+TEST(mdspan_macros, precondition_check_constexpr_invocable)
+{
+  struct fn
+  {
+    constexpr auto operator()() const
+    {
+      MDSPAN_PRECONDITION(1 + 1 == 2);
```
crtrott:
```suggestion
      MDSPAN_IMPL_PRECONDITION(1 + 1 == 2);
```

crtrott (changes requested):
This looks pretty good. Finally found sufficient time to review it. The new macros are not a bad idea, but I think they should all be IMPL macros.

crtrott:
I will rebase this and merge my suggestions.

crtrott:
@oliverlee FYI: I rebased and force pushed. I'd like to get this in before I work on more of the features for C++26.

crtrott (approved):

oliverlee:
Not sure why the death tests are failing now - they were fine a few months back.

I'll probably have time to look into it tomorrow.

crtrott:
They passed now. We might have changed Debug to Release build at some point and so the rebase made them fail.

oliverlee:
It should be possible to enable precondition checks in release mode with a preprocessor macro - I imagine this is enabled only for tests and disabled for benchmarks.

It shouldn't be too much trouble to do in a separate PR if you want to merge this PR and use it as a base for c++26 stuff.

oliverlee:
@crtrott It looks like the macro in `test/CMakeLists.txt` was missed in the rename commit.

I've created two commits - please squash or do whatever is normal for this repo.

crtrott:
Thanks! I had overlooked that. That said I am not sure it's the best way to go about this. Now the tests never take the code path where the preconditions are skipped. Maybe instead we should have a mix of debug and non-debug test configs.

crtrott:
Also I am waiting for a second review to finish (should happen latest by Monday I pinged appropriate people in my team).

oliverlee:
> That said I am not sure it's the best way to go about this. 

Same - whatever makes the most sense to you in terms of maintenance.

I've seen precondition tests defined in a separate test file. That could remove the need to conditionally define test cases with `ASSERT_DEATH` and simplify running tests with debug and non-debug configs. Would that make sense here?

crtrott:
It could make sense to separate them out. We had some platforms where the death tests never fully worked ...

oliverlee:
> It could make sense to separate them out. We had some platforms where the death tests never fully worked ...

There are only a few `DEATH_TESTS` now.

It makes sense to do it earlier rather than later if this is approach you want to pursue.

crtrott:
Yeah let's do it. Do you have time to tackle that? If not I might get around to it some point this weekend.

oliverlee:
> Yeah let's do it. Do you have time to tackle that? If not I might get around to it some point this weekend.

Maybe this afternoon? If not, also this weekend.

crtrott:
Sounds good. Also I send you an invite to our slack channel in case you are interested.

crtrott:
Note: I merged one other PR, but this one rebases cleanly on top of it.

oliverlee:
Rebased + squashed some commits + separated out the death tests. Is this what you had in mind?

crtrott:
Yeah this is great. Thanks so much!

include/experimental/__p0009_bits/macros.hpp
```diff
@@ -101,6 +103,52 @@
...
+  #else
+    #define MDSPAN_IMPL_CHECK_PRECONDITION 1
+  #endif
+#endif
+
+namespace MDSPAN_IMPL_STANDARD_NAMESPACE {
+namespace detail {
+
+template <bool check = MDSPAN_IMPL_CHECK_PRECONDITION>
+constexpr void precondition(const char* cond, const char* file, unsigned line)
```
dalg24: I think this needs to be decorated with `MDSPAN_FUNCTION`

dalg24: Did you consider tag dispatching instead ,having two non templated overloads
```C++
MDSPAN_FUNCTION constexpr void precondition(std::false_type /*checked*/, const char* /*cond*/, const char* /*file*/, unsigned /*line*/) { /* do nothing */ }
MDSPAN_FUNCTION constexpr void precondition(std::true_type /*checked*/, const char* cond, const char* file, unsigned line) { MDSPAN_IMPL_PRECONDITION_VIOLATION_HANDLER(cond, file, line); }

File: include/experimental/__p0009_bits/utility.hpp [hunk 16]
```diff
@@ -0,0 +1,62 @@
...
+} extent;
+
+constexpr struct
+{
+  template <class T, class I>
+  constexpr auto operator()(const T& x, I i) const
+  {
+    return x.stride(i);
+  }
+} stride;
```
dalg24, line 59: Where are these used?

include/experimental/__p0009_bits/macros.hpp
```diff
@@ -101,6 +103,52 @@
 #define MDSPAN_IMPL_STANDARD_NAMESPACE_STRING MDSPAN_PP_STRINGIFY(MDSPAN_IMPL_STANDARD_NAMESPACE)
 #define MDSPAN_IMPL_PROPOSED_NAMESPACE_STRING MDSPAN_PP_STRINGIFY(MDSPAN_IMPL_STANDARD_NAMESPACE) "::" MDSPAN_PP_STRINGIFY(MDSPAN_IMPL_PROPOSED_NAMESPACE)
 
+namespace MDSPAN_IMPL_STANDARD_NAMESPACE {
+namespace detail {
+
+inline void default_precondition_violation_handler(const char* cond, const char* file, unsigned line)
```
dalg24: Fine by me but just noting that the default handler would not work with CUDA/HIP

include/experimental/__p0009_bits/utility.hpp
crtrott: line 526 in layout_stride.hpp and 546 or so in extents.hpp respectively.

include/experimental/__p2642_bits/layout_padded_fwd.hpp
```diff
@@ -82,36 +83,49 @@ struct is_layout_right_padded_mapping<_Mapping,
...
+constexpr void check_padded_layout_converting_constructor_preconditions(MDSPAN_IMPL_STANDARD_NAMESPACE::detail::with_rank<1>,
+                                                                        const _OtherMapping&) {}
+template <typename _ExtentsType, typename _OtherMapping, std::size_t N>
+constexpr void check_padded_layout_converting_constructor_preconditions(MDSPAN_IMPL_STANDARD_NAMESPACE::detail::with_rank<N>,
+                                                                        const _OtherMapping &other_mapping) {
+  constexpr auto padded_stride_idx =
+    layout_padded_constants<typename _OtherMapping::layout_type,
+                            _ExtentsType>::padded_stride_idx;
+  constexpr auto extent_to_pad_idx = layout_padded_constants<typename _OtherMapping::layout_type, _ExtentsType>::extent_to_pad_idx;
+  assert(other_mapping.stride(padded_stride_idx) == other_mapping.extents().extent(extent_to_pad_idx));
```
dalg24: Should we not be using `MDSPAN_IMPL_PRECONDITION` here?

tests/CMakeLists.txt
```diff
@@ -6,14 +6,24 @@ if(MDSPAN_ENABLE_CUDA)
...
+  cmake_parse_arguments(ARGUMENT "${options}" "${one_value}" "${multi_value}" ${ARGN})
+
   if(MDSPAN_TEST_LANGUAGE)
     set_source_files_properties(${name} PROPERTIES LANGUAGE ${MDSPAN_TEST_LANGUAGE})
   endif()
   add_executable(${name} ${name}.cpp)
   target_link_libraries(${name} mdspan gtest_main)
   add_test(${name} ${name})
-endmacro()
+  set_property(TARGET ${name} PROPERTY COMPILE_WARNING_AS_ERROR ON)
```
dalg24, line 25: I am not necessarily opposed to this but IMO it does not belong to this PR

tests/test_alternate_precondition_violation_handler.cpp
```diff
@@ -0,0 +1,29 @@
+#include <stdexcept>
+
+#define MDSPAN_IMPL_PRECONDITION_VIOLATION_HANDLER(cond, file, line) \
+  do { \
+    throw std::logic_error{"precondition failure"}; \
+  } while (0);
```
dalg24, line 6: I am confused.  What is this suppose to test?
~~For one it is throwing regardless of whether the condition evaluates to false or not.~~ **EDIT: I see now that the condition is meant for debugging purposes, not for evaluation.  Sorry about the noise**
Also the `MDSPAN_IMPL_` denotes that that the macro is "private" or "internal" which indicates it is not meant as a customization point for users.

tests/test_macros.cpp
dalg24, line 1: Do we care to test that precondition checking can indeed be disabled?

include/experimental/__p0009_bits/macros.hpp
oliverlee: I might have a while back? Not sure.

But I tested it out just now, and this approach produces warnings with Clang and GCC:

Clang-18:
```
.../mdspan/include/mdspan/../experimental/__p0009_bits/macros.hpp:142:16: error: constexpr function never produces a constant expression [-Winvalid-constexpr]
  142 | constexpr void precondition(std::true_type, const char* cond, const char* file, unsigned line)
```

GCC-11:
```
.../mdspan/include/mdspan/../experimental/__p0009_bits/macros.hpp:120:81: error: call to non-‘constexpr’ function ‘void Kokkos::detail::default_precondition_violation_handler(const char*, const char*, unsigned int)’
```

include/experimental/__p0009_bits/macros.hpp
oliverlee: Please feel free to make any changes if CUDA/HIP compatibility is necessary.

include/experimental/__p2642_bits/layout_padded_fwd.hpp
oliverlee: Sure let's do that.

tests/CMakeLists.txt
oliverlee: This change is related to C++14 support to some degree - otherwise C++17 language features are still usable with some compilers when building for C++14. But I also see that the impact of this change has wider consequences.

What do you think about breaking out this change into a separate PR and rebasing this PR on top of that? I can do that later this evening - or feel free to change this PR as you see fit.

tests/test_alternate_precondition_violation_handler.cpp
oliverlee: No worries about the noise. Should there be some documentation here for future readers?

tests/test_macros.cpp
oliverlee: Yes that's probably a good idea.

I'm not sure what's the best way - possibly a new test file+target that defines a precondition handler that unconditionally throws and then setting `MDSPAN_IMPL_CHECK_PRECONDITION 0`?

include/experimental/__p0009_bits/macros.hpp
dalg24: I expect the issue is the `constexpr` specifier

dalg24: I suppose you needed it to be call it in other functions that themselves are callable in a constant expression...

tests/test_alternate_precondition_violation_handler.cpp
crtrott: The original idea was to have the macros public. But I wanted to consider that a bit more in-depth and thus said to make it private for now. We should still test the mechanism.

dalg24: Sure

include/experimental/__p0009_bits/macros.hpp
mhoemmen: `printf` works just fine with CUDA; it's HIP that's broken lol ; - )

include/experimental/__p0009_bits/macros.hpp
```diff
@@ -101,6 +103,52 @@
...
+{
+  (void)std::fprintf(::stderr, "%s:%u: precondition failure: `%s`\n", file, line, cond);
+  std::abort();
+}
```
mhoemmen: I don't know how HIP or SYCL work, but the following should be fine for CUDA.  Note that in device code, `printf` may have a performance cost even if it is never called at run time.  (For example, the compiler may need to reserve registers for it.)

```suggestion
{
#if defined(__CUDA_ARCH__)
#  if ! defined(NDEBUG)
  printf("%s:%u: precondition failure: `%s`\n", file, line, cond);
  assert(false);
#  endif
#else
  fprintf(stderr, "%s:%u: precondition failure: `%s`\n", file, line, cond);
  assert(false);
#endif 
}
```

masterleinad: `HIP` works the same as `Cuda`. For SYCL (oneAPI), you would need to use `sycl::ext::oneapi::experimental::printf`.

include/experimental/__p0009_bits/macros.hpp
dalg24: it is a `fprintf` and it is qualified so there is no way it works on CUDA.  As far as I know HIP behaves just the same as CUDA here.

tests/CMakeLists.txt
oliverlee: https://github.com/kokkos/mdspan/pull/328

include/experimental/__p0009_bits/layout_left.hpp
```diff
@@ -133,11 +134,11 @@ class layout_left::mapping {
       : __extents(__other.extents())
     {
```
nmm0, line 137: This should just be guarded by C++17. We shouldn't include the layout_padded_fwd.hpp header in C++14 mode

include/experimental/__p2642_bits/layout_padded_fwd.hpp
```diff
@@ -82,36 +83,49 @@ struct is_layout_right_padded_mapping<_Mapping,
   std::enable_if_t<std::is_same<_Mapping, typename layout_right_padded<_Mapping::padding_value>::template mapping<typename _Mapping::extents_type>>::value>>
     : std::true_type {};
 
+
```
nmm0, line 86: Let's just not include this header in C++14, since it's only used for padded layouts which require at least C++17

include/experimental/__p0009_bits/layout_right.hpp
```diff
@@ -18,8 +18,8 @@
 #include "macros.hpp"
 #include "trait_backports.hpp"
 #include "extents.hpp"
-#include <stdexcept>
 #include "layout_stride.hpp"
+#include "utility.hpp"
 #include "../__p2642_bits/layout_padded_fwd.hpp"
```
nmm0, line 24: Same here, let's guard this with `MDSPAN_HAS_CXX_17`

nmm0 (dismissed):
I'm good with almost everything, just lets keep the padded layout stuff isolated (since it requires C++17 anyway)

crtrott:
I just noticed that this requires CXX Extensions for GCC. I.e. -std=gnu++14 - it doesn't work with -std=c++14. Do you consider that acceptable?

include/experimental/__p0009_bits/layout_left.hpp
crtrott: done

oliverlee:
> I just noticed that this requires CXX Extensions for GCC. I.e. -std=gnu++14 - it doesn't work with -std=c++14. Do you consider that acceptable?

Do you know what requires extensions? I didn't intend for that and I believe I am only using -std=c++14 in projects using mdspan (however, those don't use CMake).

crtrott:
Its now fixed.

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -447,32 +451,48 @@ struct layout_stride {
...
+        if (extents().extent(r) == 0 && r != r_largest) {
+          return false;
         }
       }
+      return true;
+    }
+
+  public:
+    MDSPAN_INLINE_FUNCTION _MDSPAN_CONSTEXPR_14 bool is_exhaustive() const noexcept {
+      return is_exhaustive_impl(detail::with_rank<extents_type::rank()>{});
```
dalg24, line 495: No warning from `is_exhaustive_impl` not having the host device markups?

include/experimental/__p0009_bits/macros.hpp
```diff
@@ -18,7 +18,12 @@
 
 #include "config.hpp"
 
+#include <cstdio>
+#include <cstdlib>
 #include <type_traits> // std::is_void
+#if defined(_MDSPAN_HAS_CUDA) || defined(_MDSPAN_HAS_HIP) || defined(_MDSPAN_HAS_SYCL)
+#include "assert.h"
```
dalg24, line 25: Please confirm that you mean to include this header and not `<cassert>`

include/experimental/__p0009_bits/macros.hpp
```diff
@@ -101,6 +106,69 @@
...
 #define MDSPAN_IMPL_PROPOSED_NAMESPACE_STRING MDSPAN_PP_STRINGIFY(MDSPAN_IMPL_STANDARD_NAMESPACE) "::" MDSPAN_PP_STRINGIFY(MDSPAN_IMPL_PROPOSED_NAMESPACE)
 
+namespace MDSPAN_IMPL_STANDARD_NAMESPACE {
+namespace detail {
+
+#if defined(_MDSPAN_HAS_CUDA) || defined(_MDSPAN_HAS_HIP)
+MDSPAN_FUNCTION inline void default_precondition_violation_handler(const char* cond, const char* file, unsigned line)
+{
+  printf("%s:%u: precondition failure: `%s`\n", file, line, cond);
+  assert(0);
```
dalg24, line 116: Please confirm that you did not see issue with Clang as the CUDA compiler.

tests/CMakeLists.txt
dalg24: Christian did you mean to keep this?

tests/test_layout_preconditions.cpp
```diff
@@ -0,0 +1,32 @@
...
+#include <gtest/gtest.h>
+
+TEST(TestConvertingConstructionFromLayoutStride, precondition_failure) {
+  using E = Kokkos::extents<size_t, 2, 2>;
+
+  const auto stride = Kokkos::layout_stride::mapping<E>{E{}, std::array<size_t, 2>{2, 8}};
+
+#if defined(_MDSPAN_HAS_CUDA) || defined(_MDSPAN_HAS_HIP) || defined(_MDSPAN_HAS_SYCL)
+  ASSERT_DEATH(Kokkos::layout_left::mapping<E>{stride}, "");
+  ASSERT_DEATH(Kokkos::layout_right::mapping<E>{stride}, "");
```
dalg24, line 27: Wait why don't we special cases these?  Is the error message lost?

tests/test_mdspan_size.cpp
```diff
@@ -55,6 +55,7 @@ void test_mdspan_size(std::vector<char>& storage, Extents&& e)
 
 TEST(TestMdspan, MdspanSizeReturnTypeAndPrecondition)
 {
+  (void) test_info_;
```
dalg24, line 58: Yuk

tests/test_macros.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include <gtest/gtest.h>
+
+TEST(mdspan_macros, precondition_violation)
+{
+#if defined(_MDSPAN_HAS_CUDA) || defined(_MDSPAN_HAS_HIP) || defined(_MDSPAN_HAS_SYCL)
+  constexpr auto msg = "";
+#else
+  constexpr auto msg = "hello, world!";
+#endif
+
```
dalg24, line 11: I would have done
```c++
#if MDSPAN_IMPL_CHECK_PRECONDITION
ASSERT_DEATH(MDSPAN_IMPL_PRECONDITION(false and "hello, world!"), msg);
#else
ASSERT_NO_THROW(MDSPAN_IMPL_PRECONDITION(false and "hello, world!"));
#endif
```
instead of messing with the test success criterion in the cmake file.

tests/test_layout_preconditions.cpp
crtrott: Yeah because we don't use all the __assertfail crap.

tests/test_macros.cpp
dalg24: Please ignore.   This doesn't apply here.  I is only for `test_precondition_checks_can_be_disabled` Sorry about the noise.

include/experimental/__p0009_bits/macros.hpp
crtrott: I couldn't find a combo which fails. For Kokkos we anyway gonna replace the thing with Kokkos::abort.

include/experimental/__p0009_bits/layout_stride.hpp
crtrott: we are currently using constexpr-experimental stuff for mdspan, we will fix that later.

dalg24: fine

include/experimental/__p0009_bits/macros.hpp
crtrott: Yeah I couldn't find a variant which fails and we will use in Kokkos Kokkos::abort anyway.

dalg24 (approved):
Good enough.
Thank you Oliver.

PR #317, opened 2024-02-27
Fix #316: replace `throw` with `std::abort` in `layout_right.hpp`

### LINKED ISSUE #316: Cannot build with exceptions disabled
There is a single `throw` in the source in `layout_right.hpp`: https://github.com/kokkos/mdspan/blob/0e6a69dfe045acbb623003588a4aff844ea4b276/include/experimental/__p0009_bits/layout_right.hpp#L164-L167

This prevents the library from compiling with `-fno-exceptions`, typical for embedded environments. The comment states that this indeed causes a terminate anyway, it would probably be better to use  `std::terminate` or `std::abort` directly.

Interestingly, the matching `throw` within `layout_left.hpp` must have been refactored to that end some time between `v0.6.0` and `stable`:
https://github.com/kokkos/mdspan/blob/879869d6740c47d40c02754cc05e241c1c66520f/include/experimental/__p0009_bits/layout_left.hpp#L164-L165

So this looks like an unfinished refactor to me.

burnpanck:
See #316.

### DIFF
--- a/include/experimental/__p0009_bits/layout_right.hpp
+++ b/include/experimental/__p0009_bits/layout_right.hpp
@@ -160,11 +160,10 @@ class layout_right::mapping {
        #if !defined(_MDSPAN_HAS_CUDA) && !defined(_MDSPAN_HAS_HIP) && !defined(NDEBUG)
        if constexpr (extents_type::rank() > 0) {
          index_type stride = 1;
+         using common_t = std::common_type_t<index_type, typename OtherExtents::index_type>;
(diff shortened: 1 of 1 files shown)


burnpanck:
Sorry for the amending/force-pushing - I should really stop pushing PR before I have tested the fix; even when it's just a single-line fix.

mhoemmen (approved):
Thanks for the fix! : - ) 

For other readers: `layout_left.hpp` already has the `std::abort` instead of a `throw`.

PR #323, opened 2024-04-02
Fix LWG Issue 4060 (fix submdspan to prevent invalid pointer creation)

mhoemmen:
Link to issue: https://cplusplus.github.io/LWG/issue4060

1. Fix submdspan to prevent invalid pointer creation
2. Implement bounds checking for the various layouts, so we can actually detect the issue

I aimed for C++14 backwards compatibility.  Item (2) above is not strictly necessary to fix the issue, but it does help with finding the issue in the first place.

Bounds checking is only enabled in a debug build (if `NDEBUG` is not defined).  It involves some fold expressions over `assert` expressions.  I'm not attached to the use of `assert` for reporting errors.

### DIFF
--- a/include/experimental/__p0009_bits/extents.hpp
+++ b/include/experimental/__p0009_bits/extents.hpp
@@ -656,7 +656,7 @@ check_one_index(InputIndex user_index,
                 ExtentsIndexType current_extent)
 {
   check_lower_bound(user_index, current_extent,
-    std::bool_constant<std::is_signed_v<ExtentsIndexType>>{});
+    std::integral_constant<bool, std::is_signed_v<ExtentsIndexType>>{});
   check_upper_bound(user_index, current_extent);
 }
  
--- a/include/experimental/__p0009_bits/extents.hpp
+++ b/include/experimental/__p0009_bits/extents.hpp
@@ -615,50 +615,58 @@ static
 #endif
 constexpr bool __is_extents_v = __is_extents<T>::value;
 
-template<class UserIndexType, class IndexType>
-constexpr void check_lower_bound(const UserIndexType& user_index,
-                                 const IndexType& /* current_extent */,
-                                 std::true_type) /* is_signed */
+template<class InputIndexType, class ExtentsIndexType>
+MDSPAN_INLINE_FUNCTION
+constexpr void
+check_lower_bound(InputIndexType user_index,
+                  ExtentsIndexType /* current_extent */,
+                  std::true_type /* is_signed */)
 {
   (void) user_index; // prevent unused variable warning
 #ifdef _MDSPAN_DEBUG
-  assert(static_cast<IndexType>(user_index) >= 0);
+  assert(static_cast<ExtentsIndexType>(user_index) >= 0);
 #endif
 }
 
-template<class UserIndexType, class IndexType>
+template<class InputIndexType, class ExtentsIndexType>
+MDSPAN_INLINE_FUNCTION
 constexpr void
-check_lower_bound(const UserIndexType& /* user_index */,
-                  const IndexType& /* current_extent */,
-                  std::false_type) /* is_signed */
+check_lower_bound(InputIndexType /* user_index */,
+                  ExtentsIndexType /* current_extent */,
+                  std::false_type /* is_signed */)
 {}
 
-template<class UserIndexType, class IndexType>
+template<class InputIndexType, class ExtentsIndexType>
+MDSPAN_INLINE_FUNCTION
 constexpr void
-check_upper_bound(const UserIndexType& user_index,
-                  const IndexType& current_extent)
+check_upper_bound(InputIndexType user_index,
+                  ExtentsIndexType current_extent)
 {
   (void) user_index; // prevent unused variable warnings
   (void) current_extent;
 #ifdef _MDSPAN_DEBUG
-  assert(static_cast<IndexType>(user_index) < current_extent);
+  assert(static_cast<ExtentsIndexType>(user_index) < current_extent);
 #endif
 }
 
-template<class InputIndex, class IndexType>
+template<class InputIndex, class ExtentsIndexType>
+MDSPAN_INLINE_FUNCTION
 constexpr void
-check_one_index(const InputIndex& user_index,
-                const IndexType& current_extent)
+check_one_index(InputIndex user_index,
+                ExtentsIndexType current_extent)
 {
   check_lower_bound(user_index, current_extent,
-                    std::bool_constant<std::is_signed_v<IndexType>>{});
+    std::bool_constant<std::is_signed_v<ExtentsIndexType>>{});
   check_upper_bound(user_index, current_extent);
 }
  
-template<class ... Indices, class Extents, size_t ... RankIndices>
+template<size_t ... RankIndices,
+         class ExtentsIndexType, size_t ... Exts,
+         class ... Indices>
+MDSPAN_INLINE_FUNCTION
 constexpr void
 check_all_indices_helper(std::index_sequence<RankIndices...>,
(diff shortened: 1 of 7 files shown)


include/experimental/__p0009_bits/extents.hpp
```diff
@@ -614,5 +615,65 @@ static
...
+template<class UserIndexType, class IndexType>
+constexpr void check_lower_bound(const UserIndexType& user_index,
+                                 const IndexType& /* current_extent */,
+                                 std::true_type) /* is_signed */
+{
+#if defined(NDEBUG)
+  (void) user_index;
+#else
+  assert(static_cast<IndexType>(user_index) >= 0);
+#endif
```
dalg24:
```suggestion
#if defined(NDEBUG)
  (void) user_index;
#endif
  assert(static_cast<IndexType>(user_index) >= 0);
```
same comment everywhere.
(Rational: to mitigate the risk for a misspelled macro)
Alternatively drop the macro guards altogether and just unconditionally cast to void.

tests/test_submdspan.cpp
```diff
@@ -275,3 +275,77 @@ TYPED_TEST(TestSubMDSpan, submdspan_return_type) {
...
 }
+
+TEST(TestSubmdspanIssue4060, Rank1) {
+  auto x = std::array<int, 3>{};
+  auto A = Kokkos::mdspan{x.data(), Kokkos::extents{3}};
+  auto B = Kokkos::submdspan(A, std::tuple{3, 3});
+
+  ASSERT_EQ(B.rank(), 1u);
+  EXPECT_EQ(B.extent(0), 0);
+  EXPECT_EQ(B.data_handle(), x.data() + 3);
```
dalg24: Wanna use `required_span_size` here?

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -273,6 +319,18 @@ layout_stride::mapping<Extents>::submdspan_mapping_impl(
       std::index_sequence<>(),
       slices...);
   using dst_mapping_t = typename layout_stride::template mapping<dst_ext_t>;
+
+  // Figure out if any slice's lower bound equals the corresponding extent.
+  // If so, bypass evaluating the layout mapping.  This fixes LWG Issue 4060.
+  const bool out_of_bounds =
+    detail::any_slice_out_of_bounds(this->extents(),
+      std::forward<SliceSpecifiers>(slices)...);
```
dalg24: What does the forward do here?  The argument are passed by value and they are used again multiple times below.
Same comment for all other occurrences.

include/experimental/__p0009_bits/layout_left.hpp
```diff
@@ -194,6 +194,9 @@ class layout_left::mapping {
     )
     _MDSPAN_HOST_DEVICE
     constexpr index_type operator()(Indices... idxs) const noexcept {
+#if ! defined(NDEBUG)
+      detail::check_all_indices(this->extents(), idxs...);
```
dalg24, line 198: I think that might be an issue.
`check_all_indices` does not have the host device annotations.
Also I think C assert is not portable on the device side. Like I am pretty sure it breaks for Clang+CUDA.

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -614,5 +615,65 @@ static
 #endif
 constexpr bool __is_extents_v = __is_extents<T>::value;
 
+template<class UserIndexType, class IndexType>
+constexpr void check_lower_bound(const UserIndexType& user_index,
+                                 const IndexType& /* current_extent */,
```
dalg24: I'd pass these by value because that is what the access operators do.

include/experimental/__p0009_bits/extents.hpp
mhoemmen: Good point -- I could also use `[[maybe_unused]]` here but I didn't want to make life harder for users who want a C++14 back-port.

include/experimental/__p0009_bits/layout_left.hpp
mhoemmen: @dalg24 wrote:
> Also I think C assert is not portable on the device side. Like I am pretty sure it breaks for Clang+CUDA.

When you say "Clang + CUDA," do you mean "using Clang to build both host and device code"?

The `assert` macro works just fine in device code if I build with NVCC as the device compiler.  We use it in CUTLASS all the time, and test regularly with the "Clang as host compiler, NVCC as device compiler" option.

Is the issue that this project must support certain other programming models where it doesn't work?  If so, then what's the recommended way to express a precondition violation?

mhoemmen: The same header (extents.hpp) uses `assert` in functions marked `MDSPAN_INLINE_FUNCTION`, but guards it with `_MDSPAN_DEBUG`.  I'll follow that pattern for now.

mhoemmen: I've guarded use of `assert` with `_MDSPAN_DEBUG`, as other uses of `assert` in the file already do.

include/experimental/__p2630_bits/submdspan_mapping.hpp
mhoemmen: I've replaced this by passing the slices by const reference instead.  Thanks!

include/experimental/__p0009_bits/layout_left.hpp
mhoemmen: I've decorated the new bounds checking functions with `MDSPAN_INLINE_FUNCTION`.

include/experimental/__p0009_bits/extents.hpp
mhoemmen: Done!

include/experimental/__p0009_bits/extents.hpp
mhoemmen: Fixed; thanks!

tests/test_submdspan.cpp
mhoemmen: Done!

include/experimental/__p0009_bits/layout_left.hpp
dalg24: > When you say "Clang + CUDA," do you mean "using Clang to build both host and device code"?

Yes.

> Is the issue that this project must support certain other programming models where it doesn't work?

Yes we will deploy in Kokkos as the underlying implementation of views.  Nic has been working on it.

I wasn't sure how we supposed to do debug assertions but it looks like you found it.

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -614,5 +615,75 @@ static
...
+}
+
+template<class InputIndex, class ExtentsIndexType>
+MDSPAN_INLINE_FUNCTION
+constexpr void
+check_one_index(InputIndex user_index,
+                ExtentsIndexType current_extent)
+{
+  check_lower_bound(user_index, current_extent,
+    std::bool_constant<std::is_signed_v<ExtentsIndexType>>{});
```
dalg24:
```suggestion
    std::integral_constant<bool, std::is_signed_v<ExtentsIndexType>>{});
```
(`bool_constant` is C++17)

dalg24 (approved):
LGTM (after you fix the bool_constant)

include/experimental/__p0009_bits/extents.hpp
mhoemmen: Good catch!  `true_type` and `false_type` appear to have been introduced in C++11, thankfully.

crtrott (approved):

PR #327, opened 2024-04-24
fix padded layout annotations and add tagged raw array constructor to `layout_stride`

### LINKED ISSUE #326: missing `__host__` `__device__` in padded layouts
Some functions need to be marked with `MDSPAN_INLINE_FUNCTION` which in cuda generate `__host__` `__device__`

nmm0:
Fixes #326

Additionally, some device code cannot use `std::array` so this adds a "non-standard" constructor to `layout_stride` that accepts a raw array to support this use case

### DIFF
--- a/include/experimental/__p0009_bits/layout_stride.hpp
+++ b/include/experimental/__p0009_bits/layout_stride.hpp
@@ -202,6 +202,12 @@ struct layout_stride {
         return __strides_storage_t{static_cast<index_type>(s[Idxs])...};
       }
 
+      template<class IntegralType>
+      MDSPAN_INLINE_FUNCTION
+      static constexpr const __strides_storage_t fill_strides(mdspan_non_standard_tag, const IntegralType (&s)[extents_type::rank()]) {
+        return __strides_storage_t{static_cast<index_type>(s[Idxs])...};
+      }
+
 #ifdef __cpp_lib_span
       template<class IntegralType>
       MDSPAN_INLINE_FUNCTION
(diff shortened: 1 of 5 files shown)


include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -309,6 +315,44 @@ struct layout_stride {
...
+#endif
+    {
+      /*
+       * TODO: check preconditions
+       * - s[i] > 0 is true for all i in the range [0, rank_ ).
+       * - REQUIRED-SPAN-SIZE(e, s) is a representable value of type index_type ([basic.fundamental]).
+       * - If rank_ is greater than 0, then there exists a permutation P of the integers in the
+       *   range [0, rank_), such that s[ pi ] >= s[ pi − 1 ] * e.extent( pi − 1 ) is true for
+       *   all i in the range [1, rank_ ), where pi is the ith element of P.
+       */
```
nmm0, line 357: This TODO is just a copy and paste from the `std::array` constructor

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -309,6 +315,44 @@ struct layout_stride {
...
+        // MSVC 19.32 does not like using index_type here, requires the typename Extents::index_type
+        // error C2641: cannot deduce template arguments for 'MDSPAN_IMPL_STANDARD_NAMESPACE::layout_stride::mapping'
+        _MDSPAN_TRAIT(std::is_convertible, const std::remove_const_t<IntegralTypes>&, typename Extents::index_type) &&
+        _MDSPAN_TRAIT(std::is_nothrow_constructible, typename Extents::index_type, const std::remove_const_t<IntegralTypes>&)
+      )
+    )
+    MDSPAN_INLINE_FUNCTION
+    constexpr
+    mapping(
+      mdspan_non_standard_tag,
```
mhoemmen, line 334: I appreciate the tag!  This will make it clear that the feature is nonstandard.

Our approach to issues like this has been to create our own copies of Standard classes and make them available in libcu++.  For example, `#include <cuda/std/array>` gets you `cuda::std::array`, which acts just like `std::array`, but has the appropriate `__device__ __host__` blessings.  On the other hand, this approach is fine, especially since the tag makes it hard to call this by accident.

mhoemmen (approved):
Thanks for this fix!

include/experimental/__p0009_bits/layout_stride.hpp
nmm0: Haha yeah Christian came up with the tag idea.

This change was necessary because of a particular use case in Kokkos proper and we can't use span yet. Maybe we should include an "array" class in mdspan in the future but at least for now I just wanted to get something that was usable since it was blocking a Kokkos feature.

include/experimental/__p0009_bits/macros.hpp
```diff
@@ -629,3 +629,8 @@ struct __bools;
 
 // </editor-fold> end Pre-C++14 constexpr }}}1
 //==============================================================================
+
+namespace MDSPAN_IMPL_STANDARD_NAMESPACE {
+constexpr struct mdspan_non_standard_tag {
+} mdspan_non_standard;
+} // namespace MDSPAN_IMPL_STANDARD_NAMESPACE
```
crtrott: This should go into the utility header introduced by the cpp14 support PR

crtrott (changes requested):
Did you ran clang-format over this thing?

crtrott (approved):

PR #328, opened 2024-05-06
treat warnings as errors in mdspan tests

oliverlee:
When compiling mdspan tests, treat compiler warnings as error. This option is *not* propagated to consumers of mdspan.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -13,6 +13,7 @@ macro(mdspan_add_test name)
   add_executable(${name} ${name}.cpp)
   target_link_libraries(${name} mdspan gtest_main)
   add_test(${name} ${name})
+  set_property(TARGET ${name} PROPERTY COMPILE_WARNING_AS_ERROR ON)
 endmacro()
 
 if(MDSPAN_USE_SYSTEM_GTEST)


tests/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ macro(mdspan_add_test name)
   add_executable(${name} ${name}.cpp)
   target_link_libraries(${name} mdspan gtest_main)
   add_test(${name} ${name})
+  set_property(TARGET ${name} PROPERTY COMPILE_WARNING_AS_ERROR ON)
```
nmm0, line 16: This isn't supported by the (admittedly quite old) minimum CMake version we require. I'd suggest an option (i.e. `MDSPAN_WARNINGS_AS_ERRORS`) that defaults to `ON` and then we use custom cmake logic to set the associated flags. Alternatively, I'd support bumping our minimum cmake version to 3.24 but not sure how others feel about that :P

mhoemmen: People may want the option to turn off warnings-as-errors for tests, especially since different compilers emit different warnings (some of which are spurious).  A CMake option sounds like something we would want anyway.

nmm0: agreed -- though I just found out with `COMPILE_WARNING_AS_ERROR` you can pass [`--compile-no-warning-as-error`](https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-compile-no-warning-as-error) to cmake to suppress it. But that's only with cmake 3.24

mhoemmen: Neat!  That sounds particularly useful, esp. for people who have to configure for MSVC but don't know what flag to use.

nmm0:
Looks like we already merged this in https://github.com/kokkos/mdspan/pull/314

So I'm closing this

PR #332, opened 2024-05-21
Fix missing restriction on ElementType

### LINKED ISSUE #330: Missing restriction on ElementType and AccessorPolicy​::​element_type being the same
According to [the standard draft](https://eel.is/c++draft/views#mdspan.mdspan.overview-2.3), [cppreference](https://en.cppreference.com/w/cpp/container/mdspan) and common sense these need to be the same, but at least when using the single header version I can instantiate e.g. `std::mdspan<float, std::dextents<int, 2>, std::layout_right, std::default_accessor<double>>` or `std::mdspan<float const, std::dextents<int, 2>, std::layout_right, std::default_accessor<float>>` without a compiler error.

See https://godbolt.org/z/nx4YE91G3

pauleonix:
Closes #330 

It needs to be the same as `AccessorPolicy::element_type`.

I used plain `std::is_same<T, U>::value` because `std::is_same_v<T, U>` is a C++17 feature and the code should be compatible with C++14.

### DIFF
--- a/include/experimental/__p0009_bits/mdspan.hpp
+++ b/include/experimental/__p0009_bits/mdspan.hpp
@@ -34,6 +34,8 @@ class mdspan
 private:
   static_assert(detail::__is_extents_v<Extents>,
                 MDSPAN_IMPL_STANDARD_NAMESPACE_STRING "::mdspan's Extents template parameter must be a specialization of " MDSPAN_IMPL_STANDARD_NAMESPACE_STRING "::extents.");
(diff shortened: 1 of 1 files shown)


dalg24 (approved):
Thanks

For reference this is indeed listed as a mandate https://eel.is/c++draft/mdspan.mdspan#overview-2.3

**edit** and now I see that you referred to that clause in your bug report :)

mhoemmen:
@pauleonix Delightful!  Thanks for the fix! and thanks @dalg24 for the review!

PR #333, opened 2024-05-22
Fix CUDA build and run issues as well as warnings

crtrott:


crtrott:
This also fixes the fact that with CUDA 12+ we did not have deduction guides ...

include/experimental/__p0009_bits/config.hpp
```diff
@@ -205,7 +205,7 @@ static_assert(_MDSPAN_CPLUSPLUS >= MDSPAN_CXX_STD_14, "mdspan requires C++14 or
 #endif
 
 #ifndef _MDSPAN_USE_CLASS_TEMPLATE_ARGUMENT_DEDUCTION
-#  if (!defined(__NVCC__) || (__CUDACC_VER_MAJOR__ >= 11 && __CUDACC_VER_MINOR__ >= 7)) && \
+#  if (!defined(__NVCC__) || (__CUDACC_VER_MAJOR__ * 100 + __CUDACC_VER_MINOR__ >= 1107)) && \
```
dalg24: No way Jose

I was going to observe that it could be confusing because we do this in Kokkos
https://github.com/kokkos/kokkos/blob/07a500982de5fd3a0876f5ef2c1a41f7130512cb/core/src/Kokkos_Macros.hpp#L110-L111
But this inconsistent with other macro guards in this project so please don't
https://github.com/kokkos/mdspan/blob/df6d31b2e6fe1269ed5770aa1f8205308b8cab2c/include/experimental/__p2630_bits/submdspan_mapping.hpp#L179

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -176,7 +177,8 @@ layout_left::mapping<Extents>::submdspan_mapping_impl(SliceSpecifiers... slices)
                                    *this, inv_map,
     // HIP needs deduction guides to have markups so we need to be explicit
     // NVCC 11.0 has a bug with deduction guide here, tested that 11.2 does not have the issue
-    #if defined(_MDSPAN_HAS_HIP) || (defined(__NVCC__) && (__CUDACC_VER_MAJOR__ * 100 + __CUDACC_VER_MINOR__ * 10) < 1120)
+    // But Clang-CUDA also doesn't accept the use of deduction guide so disable it for CUDA alltogether
+    #if defined(_MDSPAN_HAS_HIP) || defined(_MDSPAN_HAS_CUDA)
                                    std::tuple<decltype(detail::stride_of(slices))...>{detail::stride_of(slices)...})),
```
dalg24, line 182: would `make_tuple` resolve the issue?
(don't care too much)

include/experimental/__p2642_bits/layout_padded.hpp
```diff
@@ -59,6 +59,10 @@ MDSPAN_INLINE_FUNCTION constexpr size_t get_actual_static_padding_value() {
   } else {
     return dynamic_extent;
   }
+  // Missing return statement warning from NVCC
+#ifdef __NVCC__
+  return 0;
```
dalg24, line 64: can we do `__builtin_unreachable()` instead?  not sure when support was added

tests/CMakeLists.txt
```diff
@@ -59,10 +59,10 @@ mdspan_add_test(test_exhaustive_layouts)
...
 mdspan_add_test(test_layout_stride)
 if(NOT CMAKE_CXX_STANDARD STREQUAL "14")
-mdspan_add_test(test_submdspan)
-mdspan_add_test(test_submdspan_static_slice)
-mdspan_add_test(test_layout_padded_left)
-mdspan_add_test(test_layout_padded_right)
+  mdspan_add_test(test_submdspan)
+  mdspan_add_test(test_submdspan_static_slice)
+  mdspan_add_test(test_layout_padded_left)
+  mdspan_add_test(test_layout_padded_right)
```
dalg24, line 65: I am going to pretend I did not see this

tests/test_submdspan.cpp
```diff
@@ -276,6 +276,7 @@ TYPED_TEST(TestSubMDSpan, submdspan_return_type) {
   __MDSPAN_TESTS_RUN_TEST(TestFixture::run());
 }
 
+#ifdef _MDSPAN_USE_CLASS_TEMPLATE_ARGUMENT_DEDUCTION
```
dalg24, line 279: Maybe add a comment what compiler choke on `std::tuple` CTAD(?)

include/experimental/__p2642_bits/layout_padded.hpp
masterleinad: In 11.3. There are other places that don't check the version, though.

dalg24:
```
include/experimental/__p2630_bits/submdspan_extents.hpp-140-  // Even with CUDA_ARCH protection this thing warns about calling host function
include/experimental/__p2630_bits/submdspan_extents.hpp:141:  __builtin_unreachable();
include/experimental/__p2630_bits/submdspan_extents.hpp-142-#endif
--
include/experimental/__p2630_bits/submdspan_mapping.hpp-186-#if defined(__NVCC__) && !defined(__CUDA_ARCH__) && defined(__GNUC__)
include/experimental/__p2630_bits/submdspan_mapping.hpp:187:  __builtin_unreachable();
include/experimental/__p2630_bits/submdspan_mapping.hpp-188-#endif
--
include/experimental/__p2630_bits/submdspan_mapping.hpp-301-#if defined(__NVCC__) && !defined(__CUDA_ARCH__) && defined(__GNUC__)
include/experimental/__p2630_bits/submdspan_mapping.hpp:302:  __builtin_unreachable();
include/experimental/__p2630_bits/submdspan_mapping.hpp-303-#endif

File: tests/test_submdspan.cpp (hunk 5 above)
Comment (crtrott) [reply tests/test_submdspan.cpp:279]: not this is independent. The tests use deduction guides and if we don't have them we can't do it ...

File: include/experimental/__p2630_bits/submdspan_mapping.hpp (hunk 2 above)
Comment (crtrott) [reply include/experimental/__p2630_bits/submdspan_mapping.hpp:182]: I'd rather avoid even more non-marked up functions

File: include/experimental/__p2642_bits/layout_padded.hpp (hunk 3 above)
Comment (crtrott) [reply include/experimental/__p2642_bits/layout_padded.hpp:64]: Ah: but only in combo with 
```
// Suppress spurious warning with NVCC about no return statement.
// This is a known issue in NVCC and NVC++
// Depending on the CUDA and GCC version we need both the builtin
// and the diagnostic push. I tried really hard to find something shorter
// but no luck ...
#if defined __NVCC__
    #ifdef __NVCC_DIAG_PRAGMA_SUPPORT__
        #pragma nv_diagnostic push
        #pragma nv_diag_suppress = implicit_return_from_non_void_function
    #else
      #ifdef __CUDA_ARCH__
        #pragma diagnostic push
        #pragma diag_suppress implicit_return_from_non_void_function
      #endif
    #endif
#elif defined __NVCOMPILER
    #pragma    diagnostic push
    #pragma    diag_suppress = implicit_return_from_non_void_function
#endif
```
and the corresponding pragma pops

crtrott: Note in the extents case the branches return different types. Here the same type is returned no matter what which means this simpler approach works.

dalg24 (approved):
Fine

PR #336, opened 2024-05-31
Avoid use of zero-length carray

crtrott:
When compiling with pedantic there are various issues with the use of zero-length c-arrays. This PR is the result of my efforts to resolve this issue for all compilers in all C++ standard modes. In some combinations only some of the changes are needed, but to resolve all of them I needed all the changes presented here.


mhoemmen (approved):
Good comments there too! : - )

nmm0 (approved):
Looks good to me!

PR #337, opened 2024-05-31
Rework of submdspan for layout_left/right

crtrott:
This enables submdspan of layout_left and layout_right to create `layout_left_padded` and `layout_right_padded` respectively.
Actual submdspan implementation for the padded layouts will come in a subsequent PR.


crtrott:
Note this PR includes the changes from #336

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -202,9 +202,17 @@ struct layout_stride {
         return __strides_storage_t{static_cast<index_type>(s[Idxs])...};
       }
 
-      template<class IntegralType>
+      MDSPAN_TEMPLATE_REQUIRES(
+        class IntegralType,
+        // The is_convertible condition is added to make sfinae valid
```
mhoemmen: It looks like this is rebased atop PR #336.  Should PR #336 be merged first?

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -237,124 +334,141 @@ struct preserve_layout_right_mapping<std::index_sequence<Idx...>, SubRank,
...
+                        std::tuple<decltype(detail::stride_of(slices))...>{
+                            detail::stride_of(slices)...})),
+#else
+                        std::tuple{detail::stride_of(slices)...})),
+#endif
+          offset
+    };
   }
 #if defined(__NVCC__) && !defined(__CUDA_ARCH__) && defined(__GNUC__)
   __builtin_unreachable();
```
mhoemmen, line 395: Tangent: This GCC built-in works fine for us in device code (with `__CUDA_ARCH__` defined).

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -237,124 +334,141 @@ struct preserve_layout_right_mapping<std::index_sequence<Idx...>, SubRank,
...
+    auto inv_map = detail::inv_map_rank(std::integral_constant<size_t, 0>(),
+                                        std::index_sequence<>(), slices...);
+    return submdspan_mapping_result<dst_mapping_t> {
+      dst_mapping_t(dst_ext,
+                    detail::construct_sub_strides(
+                        *this, inv_map,
+// HIP needs deduction guides to have markups so we need to be explicit
+// NVCC 11.0 has a bug with deduction guide here, tested that 11.2 does not have
+// the issue But Clang-CUDA also doesn't accept the use of deduction guide so
+// disable it for CUDA alltogether
```
mhoemmen:
```suggestion
// disable it for CUDA altogether
```

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -120,85 +154,97 @@ struct preserve_layout_left_mapping<std::index_sequence<Idx...>, SubRank,
...
```
mhoemmen:
```suggestion
// disable it for CUDA altogether
```

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -120,85 +154,97 @@ struct preserve_layout_left_mapping<std::index_sequence<Idx...>, SubRank,
...
+  using deduce_layout = detail::deduce_layout_left_submapping<
+      typename dst_ext_t::index_type, dst_ext_t::rank(),
+      decltype(std::make_index_sequence<src_ext_t::rank()>()),
+      SliceSpecifiers...>;
+
+  using dst_layout_t = std::conditional_t<
+      deduce_layout::layout_left_value, layout_left,
+      std::conditional_t<
+          deduce_layout::layout_left_padded_value,
+          MDSPAN_IMPL_PROPOSED_NAMESPACE::layout_left_padded<dynamic_extent>,
```
mhoemmen, line 209: I might be missing something, but does this not implement the `layout_left_padded` with static padding stride case?  That should be possible, as in the following example.

```c++
auto mapping = layout_left::mapping<extents<int, 10, 10>>{};
auto x_storage = std::array<float, 100>{};
auto x = mdspan{x_storage.data(), mapping};
auto slice = strided_slice{
  .offset=0,
  .extent=integral_constant<int, 5>{},
  .stride=integral_constant<int, 1>{}};
// Should be `layout_left_padded<10>::mapping<extents<int, 5, 5>>`
auto x_sub = submdspan(x, strided_slice, full_extent);
```

Does the wording not permit this, or is that just future work?  Totally OK if so.

mhoemmen (approved):

include/experimental/__p2630_bits/submdspan_mapping.hpp
crtrott: The standard actually does not allow this. I.e. according to the standard you get layout_stride. I am not sure I want to change that, i.e. explicitly treating the strided_slice with integral_constant 1 for stride like a pair range. I mean we could if we really wanted to, not sure it wins us much. And it might be weird in generic code to get layout_stride vs layout_left or so back depending on an integral constant. In all other cases the layout_flavor you get really just depends on the slice flavors you have, not the particular type of slice.

That said: this PR always returns dynamic_extent for the padding value when it returns padded layouts. I am fixing that in the follow up which actually introduces submdspan for padded layouts. That anyway requires a bunch of more extensive new tests.

include/experimental/__p0009_bits/layout_stride.hpp
crtrott: Yeah already done.

include/experimental/__p2630_bits/submdspan_mapping.hpp
mhoemmen: I see -- so the thing we still need to implement is https://eel.is/c++draft/views.multidim#mdspan.sub.map.left-1.4.5 , that defines `S_static` as "the product of all values `static_extent(k)` for `k` in the range [0, u + 1)" in case all the extents in the range are static?

crtrott: Yes, I am doing it as part of the padded layouts submdspan. Should be done later today or tomorrow.

crtrott: The nice thing is that our tests are already comprehensive enough to cover those cases (i.e. getting a static stride and getting a dynamic stride in various cases).

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -93,25 +81,71 @@ construct_sub_strides(const SrcMapping &src_mapping,
...
+
+template <class IndexType, size_t SubRank, size_t... Idx,
+          class... SliceSpecifiers>
+struct deduce_layout_left_submapping<
+    IndexType, SubRank, std::index_sequence<Idx...>, SliceSpecifiers...> {
+
+  using CountRange = index_sequence_scan_impl<
+      0, (std::is_convertible_v<SliceSpecifiers, IndexType> ? 0 : 1)...>;
+  //__static_partial_sums<!std::is_convertible_v<SliceSpecifiers,
+  // IndexType>...>;
```
nmm0:
```suggestion
```

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -93,25 +81,71 @@ construct_sub_strides(const SrcMapping &src_mapping,
...
+template <class IndexType, size_t SubRank, class IndexSequence,
+          class... SliceSpecifiers>
+struct deduce_layout_left_submapping;
+
+template <class IndexType, size_t SubRank, size_t... Idx,
+          class... SliceSpecifiers>
+struct deduce_layout_left_submapping<
+    IndexType, SubRank, std::index_sequence<Idx...>, SliceSpecifiers...> {
+
+  using CountRange = index_sequence_scan_impl<
```
nmm0:
```suggestion
  using count_range = index_sequence_scan_impl<
```

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -93,25 +81,71 @@ construct_sub_strides(const SrcMapping &src_mapping,
...
+      0, (std::is_convertible_v<SliceSpecifiers, IndexType> ? 0 : 1)...>;
+  //__static_partial_sums<!std::is_convertible_v<SliceSpecifiers,
+  // IndexType>...>;
+  constexpr static int NumGaps =
```
nmm0:
```suggestion
  constexpr static int num_gaps =
```

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -93,25 +81,71 @@ construct_sub_strides(const SrcMapping &src_mapping,
...
+      (((Idx > 0 && CountRange::get(Idx) == 1 &&
+         std::is_convertible_v<SliceSpecifiers, IndexType>)
+            ? 1
+            : 0) +
+       ... + 0);
+
+  constexpr static bool layout_left_value =
```
nmm0: This (and `layout_left_padded_value`) are really difficult to parse because it's a huge boolean statement. Could we instead initialize them with evaluating a constexpr function? It would be a lot easier to maintain I think because we could use natural control flow and `if constexpr`

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -93,25 +81,71 @@ construct_sub_strides(const SrcMapping &src_mapping,
...
-          ((Idx > SubRank - 1) || // these are only integral slice specifiers
-           (std::is_same_v<SliceSpecifiers, full_extent_t>) ||
-           ((Idx == SubRank - 1) &&
-            std::is_convertible_v<SliceSpecifiers, std::tuple<size_t, size_t>>)) &&
+          // leftmost must be range
+          ((Idx == 0 &&
+            (std::is_same_v<SliceSpecifiers, full_extent_t> ||
+             std::is_convertible_v<SliceSpecifiers,
+                                   std::tuple<IndexType, IndexType>>)) ||
+           (Idx > 0 && Idx <= NumGaps && CountRange::get(Idx) == 1) ||
```
nmm0: This would be greatly simplified with `if constexpr` in an initialization function

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -93,25 +81,71 @@ construct_sub_strides(const SrcMapping &src_mapping,
...
+       ((Idx > 0 || std::is_same_v<SliceSpecifiers, full_extent_t> ||
+         std::is_convertible_v<SliceSpecifiers,
+                               std::tuple<IndexType, IndexType>>)&&...)) ||
+      // Preserve if leftmost SubRank-1 slices are full_extent_t and
+      // the leftmost SubRank slices are ranges
+      (((Idx < SubRank - 1 && std::is_same_v<SliceSpecifiers, full_extent_t>) ||
+        (Idx == SubRank - 1 &&
+         (std::is_same_v<SliceSpecifiers, full_extent_t> ||
+          std::is_convertible_v<SliceSpecifiers,
+                                std::tuple<IndexType, IndexType>>)) ||
```
nmm0: I think readability would be enhanced if you had something like `is_range_specifier(SliceSpecifier)` since this is repeated a few times

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -93,25 +81,71 @@ construct_sub_strides(const SrcMapping &src_mapping,
...
+
+  constexpr static bool layout_left_padded_value =
+      // Use layout_left_padded for rank 0
       (SubRank == 0) ||
+      // Use layout_left_padded for rank 1 if leftmost slice specifier is range
+      // like
+      ((SubRank == 1) &&
+       ((Idx > 0 || std::is_same_v<SliceSpecifiers, full_extent_t> ||
+         std::is_convertible_v<SliceSpecifiers,
+                               std::tuple<IndexType, IndexType>>)&&...)) ||
```
nmm0: Are these cases actually considered? Aren't we selecting `layout_left` if these conditions are true

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -207,28 +253,79 @@ layout_left::mapping<Extents>::submdspan_mapping_impl(SliceSpecifiers... slices)
...
+          class... SliceSpecifiers>
+struct deduce_layout_right_submapping;
+
+template <class IndexType, size_t SubRank, size_t... Idx,
+          class... SliceSpecifiers>
+struct deduce_layout_right_submapping<
+    IndexType, SubRank, std::index_sequence<Idx...>, SliceSpecifiers...> {
+
+  static constexpr size_t Rank = sizeof...(Idx);
+  using CountRange = index_sequence_scan_impl<
```
nmm0: Same comments as layout_left above

nmm0 (commented):
I did a first pass. The logic is a bit hard to follow and took me a while to get. That's partly because the slicing logic is complex but also because of the way the logic is expressed; I think it would be better to write some compile-time functions to initialize the variables that decide which layout type to use.

Also might even help readability to have an enum instead of two booleans and use that to select the type from a specialization

include/experimental/__p2630_bits/submdspan_mapping.hpp
crtrott: I think I intercept them early but in principle layout_left_padded would still be valid.

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -93,268 +121,332 @@ construct_sub_strides(const SrcMapping &src_mapping,
...
-           ((Idx == SubRank - 1) &&
-            std::is_convertible_v<SliceSpecifiers, std::tuple<size_t, size_t>>)) &&
-          ...);
+template <class IndexType, size_t SubRank, size_t... Idx,
+          class... SliceSpecifiers>
+struct deduce_layout_left_submapping<
+    IndexType, SubRank, std::index_sequence<Idx...>, SliceSpecifiers...> {
+
+  using count_range = index_sequence_scan_impl<
+      0, (is_index_slice_v<SliceSpecifiers, IndexType> ? 0 : 1)...>;
```
nmm0, line 134:
```suggestion
      0, (is_range_slice_v<SliceSpecifiers, IndexType> ? 1 : 0)...>;
```
Might be clearer

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -93,268 +121,332 @@ construct_sub_strides(const SrcMapping &src_mapping,
...
 
   // figure out sub layout type
-  constexpr bool preserve_layout = detail::preserve_layout_left_mapping<
-      decltype(std::make_index_sequence<src_ext_t::rank()>()), dst_ext_t::rank(),
-      SliceSpecifiers...>::value;
-  using dst_layout_t =
-      std::conditional_t<preserve_layout, layout_left, layout_stride>;
+  using deduce_layout = detail::deduce_layout_left_submapping<
+      typename dst_ext_t::index_type, dst_ext_t::rank(),
+      decltype(std::make_index_sequence<src_ext_t::rank()>()),
```
nmm0: Could just be:

```suggestion
      std::make_index_sequence<src_ext_t::rank()>,
```

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -93,268 +121,332 @@ construct_sub_strides(const SrcMapping &src_mapping,
...
-          ...);
+template <class IndexType, size_t SubRank, size_t... Idx,
+          class... SliceSpecifiers>
+struct deduce_layout_left_submapping<
+    IndexType, SubRank, std::index_sequence<Idx...>, SliceSpecifiers...> {
+
+  using count_range = index_sequence_scan_impl<
+      0, (is_index_slice_v<SliceSpecifiers, IndexType> ? 0 : 1)...>;
+
+  constexpr static int num_gaps =
```
nmm0: Suggestion: for me it's more intuitive to think of this as a "first gap length" rather than num_gaps. I.e. You would assume the number of gaps for `RIIIIRIIIR` would be 2 since there are two "gaps" of single indices, but in reality it is 4 as the length of the first of the two gaps.

tests/test_submdspan.cpp
```diff
@@ -145,20 +145,22 @@ using submdspan_test_types =
     , std::tuple<Kokkos::layout_right, Kokkos::layout_right, Kokkos::extents<size_t,6,4,5,6,7,8>,  args_t<6,4,5,6,7,8>, Kokkos::extents<size_t,dyn,7,8>, int, int, int, std::pair<int,int>, Kokkos::full_extent_t, Kokkos::full_extent_t>
     , std::tuple<Kokkos::layout_right, Kokkos::layout_right, Kokkos::extents<size_t,6,4,5,6,7,8>,  args_t<6,4,5,6,7,8>, Kokkos::extents<size_t,dyn,8>, int, int, int, int, std::pair<int,int>, Kokkos::full_extent_t>
     , std::tuple<Kokkos::layout_right, Kokkos::layout_right, Kokkos::extents<size_t,6,4,5,6,7,8>,  args_t<6,4,5,6,7,8>, Kokkos::extents<size_t,8>, int, int, int, int, int, Kokkos::full_extent_t>
+    // LayoutLeft to layout_left_padded
+    , std::tuple<Kokkos::layout_left, Kokkos::Experimental::layout_left_padded<Kokkos::dynamic_extent>,  Kokkos::dextents<size_t,2>,          args_t<10,20>,       Kokkos::dextents<size_t,2>, std::pair<int,int>, Kokkos::full_extent_t>
```
nmm0: Can we add tests for `RIF`, `RIFR`, and `RIFRI`?

tests/test_submdspan.cpp
```diff
@@ -145,20 +145,22 @@ using submdspan_test_types =
...
     , std::tuple<Kokkos::layout_left, Kokkos::layout_stride,  Kokkos::dextents<size_t,2>,          args_t<10,20>,       Kokkos::dextents<size_t,2>, Kokkos::strided_slice<int,int,int>, std::pair<int,int>>
     , std::tuple<Kokkos::layout_left, Kokkos::layout_stride,  Kokkos::dextents<size_t,2>,          args_t<10,20>,       Kokkos::dextents<size_t,2>, Kokkos::strided_slice<int,int,int>, Kokkos::strided_slice<int,int,int>>
     , std::tuple<Kokkos::layout_left, Kokkos::layout_stride,  Kokkos::extents<size_t,6,4,5,6,7,8>, args_t<6,4,5,6,7,8>, Kokkos::extents<size_t,6,dyn,8>, Kokkos::full_extent_t, int, std::pair<int,int>, int, int, Kokkos::full_extent_t>
     , std::tuple<Kokkos::layout_left, Kokkos::layout_stride,  Kokkos::extents<size_t,6,4,5,6,7,8>, args_t<6,4,5,6,7,8>, Kokkos::extents<size_t,4,dyn,7>, int, Kokkos::full_extent_t, std::pair<int,int>, int, Kokkos::full_extent_t, int>
+    // layout_right to layout_right_padded
+    , std::tuple<Kokkos::layout_right, Kokkos::Experimental::layout_right_padded<Kokkos::dynamic_extent>, Kokkos::dextents<size_t,2>, args_t<10,20>, Kokkos::dextents<size_t,2>, Kokkos::full_extent_t, std::pair<int,int>>
```
nmm0, line 162: Can we add tests for `FIR`, `RFIR`, and `IRFIR`?

nmm0 (commented):
Just a couple more minor things then I think I'm good with it!

include/experimental/__p2630_bits/submdspan_mapping.hpp
crtrott: No, there are slice specifiers which are neither (strided)

nmm0: I guess we should rename this to `count_non_index` or `count_range_or_strided` then

include/experimental/__p2630_bits/submdspan_mapping.hpp
crtrott: done

include/experimental/__p2630_bits/submdspan_mapping.hpp
crtrott: used `gap_len`

tests/test_submdspan.cpp
crtrott: done

tests/test_submdspan.cpp
crtrott: done

nmm0 (approved):
Looks good to me!

ndellingwood:
Does this require a changelog entry for 4.4?

ndellingwood:
Oop sorry for noise about changelog, saw this listed in the Kokkos Planning board

PR #340, opened 2024-06-04
Fix warnings (shadow and conversion)

crtrott:
The extent and stride thing shadow something new in utility.hpp

### DIFF
--- a/include/experimental/__p0009_bits/layout_stride.hpp
+++ b/include/experimental/__p0009_bits/layout_stride.hpp
@@ -651,15 +651,15 @@ constexpr void validate_strides(with_rank<N>, Layout, const Extents& ext, const 
 
   constexpr auto is_left = std::is_same<Layout, layout_left>::value;
 
-  typename Extents::index_type stride = 1;
+  typename Extents::index_type expected_stride = 1;
 
   for (std::size_t r = 0; r < N; r++) {
     const std::size_t s = is_left ? r : N - 1 - r;
 
-    MDSPAN_IMPL_PRECONDITION(common_integral_compare(stride, other.stride(s))
+    MDSPAN_IMPL_PRECONDITION(common_integral_compare(expected_stride, other.stride(s))
                              and "invalid strides for layout_{left,right}");
 
-    stride *= ext.extent(s);
+    expected_stride *= ext.extent(s);
   }
 }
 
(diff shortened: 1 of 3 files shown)


include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -651,15 +651,15 @@ constexpr void validate_strides(with_rank<N>, Layout, const Extents& ext, const
 
   constexpr auto is_left = std::is_same<Layout, layout_left>::value;
 
-  typename Extents::index_type stride = 1;
+  typename Extents::index_type stride_ = 1;
```
mhoemmen: Should we consider avoiding use of trailing underscore for variables that aren't nonpublic data members of the class?

```suggestion
  typename Extents::index_type the_stride = 1;
```

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -651,15 +651,15 @@ constexpr void validate_strides(with_rank<N>, Layout, const Extents& ext, const
...
 
   for (std::size_t r = 0; r < N; r++) {
     const std::size_t s = is_left ? r : N - 1 - r;
 
-    MDSPAN_IMPL_PRECONDITION(common_integral_compare(stride, other.stride(s))
+    MDSPAN_IMPL_PRECONDITION(common_integral_compare(stride_, other.stride(s))
```
mhoemmen:
```suggestion
    MDSPAN_IMPL_PRECONDITION(common_integral_compare(the_stride, other.stride(s))
```

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -651,15 +651,15 @@ constexpr void validate_strides(with_rank<N>, Layout, const Extents& ext, const
...
                              and "invalid strides for layout_{left,right}");
 
-    stride *= ext.extent(s);
+    stride_ *= ext.extent(s);
```
mhoemmen:
```suggestion
    the_stride *= ext.extent(s);
```

mhoemmen (approved):
Thanks for the fixes! : - )

include/experimental/__p0009_bits/layout_stride.hpp
crtrott: renamed to `expected_stride`

nmm0:
I don't think our CI is actually testing Wshadow

masterleinad:
> I don't think our CI is actually testing Wshadow

Looks like the GitHub Workflows don't test with any warning flags and https://github.com/kokkos/mdspan/blob/stable/README.md#running-tests mentions that a lot of the configurations would not build with warnings as errors (depending on the C++ standard).

crtrott:
> I don't think our CI is actually testing Wshadow

I know. But we can touch CI separately.

nmm0 (approved):
Looks good to me!

PR #341, opened 2024-06-04
Update README.md

### LINKED ISSUE #339: What targets are defined by find_package( mdspan )?
mdspan comes with `.cmake` files, but I'm not good enough at cmake exegesis to figure out what targets they define. Can someone elucidate?

VictorEijkhout:
CMake discoverability

Fixes #339

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,25 @@ Note: There is a tag mdspan-0.4.0 which reflects the status of P0009 before
 * renaming `pointer`, `data`, `is_contiguous` and `is_always_contiguous`; and before
 * renaming `size_type` to `index_type` and introducing a new `size_type = make_unsigned_t<index_type>` alias.
 
+Building code with `mdspan`
+---------------------------
+
+mdspan is discoverable by CMake:
+
+```
+find_package( mdspan REQUIRED )
+target_link_libraries( myproject PUBLIC std::mdspan )
+```
+
+Your code can then have
+```
+#include "mdspan/mdspan.hpp"
+namespace md = Kokkos;
+...
+md::mdspan
+```
+
+
 Using `mdspan`
 --------------
 


mhoemmen (approved):
Thanks @VictorEijkhout for contributing this!  Looks good to me!

@crtrott What do you think?

mhoemmen:
@VictorEijkhout Btw I took the liberty of adding a "Fixes #339" message to your PR description, so that merging this PR will automatically close the issue.

crtrott (changes requested):
I had yesterday a discussion with Damien on this. This PR points something unfortunate out: I am not sure that the target being called std::mdspan is a very fortunate choice considering that mdspan is in the standard, and part of LLVM for example. We were wondering if we should rename the target (with providing an alias as std::mdspan for a while).

VictorEijkhout:
@crtrott I had the same thought. Especially since in the program it's still `Kokkos::mdspan`.

README.md
```diff
@@ -13,6 +13,25 @@ Note: There is a tag mdspan-0.4.0 which reflects the status of P0009 before
 * renaming `pointer`, `data`, `is_contiguous` and `is_always_contiguous`; and before
 * renaming `size_type` to `index_type` and introducing a new `size_type = make_unsigned_t<index_type>` alias.
 
+Building code with `mdspan`
```
nmm0, line 16: I think we should move this under the "Using mdspan" subheading and retitle this something like "Cmake `find_package`" or something like that

README.md
```diff
@@ -13,6 +13,25 @@ Note: There is a tag mdspan-0.4.0 which reflects the status of P0009 before
...
+---------------------------
+
+mdspan is discoverable by CMake:
+
+```
```
nmm0, line 21:
```suggestion
```cmake
```

README.md
```diff
@@ -13,6 +13,25 @@ Note: There is a tag mdspan-0.4.0 which reflects the status of P0009 before
...
+find_package( mdspan REQUIRED )
+target_link_libraries( myproject PUBLIC std::mdspan )
+```
+
+Your code can then have
+```
+#include "mdspan/mdspan.hpp"
+namespace md = Kokkos;
+...
+md::mdspan
```
nmm0, line 31: I think I would just have the include here (I personally prefer with angle brackets, but I suppose it's a matter of taste)

```suggestion
```cpp
#include <mdspan/mdspan.hpp>
```

I think we could probably leave out the namespace alias here

nmm0 (commented):
Thanks for submitting this! I just have a couple of suggestions

crtrott:
We should take this over and change what we name our cmake targets

PR #342, opened 2024-06-10
Improve submdspan testing

crtrott:
This now tests that the elements of the submdspan point to the correct element of the src. That caught a mistake in the layout_foo test layout.

Also turn of bracket operator for ICPX, which resulted in compiler crashes in the submdspan test. Note: we could upgrade to 2024.1 which works.


nmm0 (approved):
Looks good to me!

PR #344, opened 2024-06-14
Replace spelled out logical operators: breaks Kokkos windows tests

crtrott:
Errors such as:
```
D:\a\kokkos\kokkos\tpls\mdspan\include\experimental\__p0009_bits\utility.hpp(19): error : expected a "," [D:\a\kokkos\kokkos\build\core\src\kokkoscore.vcxproj]
      static_assert(std::is_integral<I1>::value and
```

### DIFF
--- a/include/experimental/__p0009_bits/layout_stride.hpp
+++ b/include/experimental/__p0009_bits/layout_stride.hpp
@@ -644,7 +644,7 @@ template <std::size_t N, class Layout, class Extents, class Mapping>
 constexpr void validate_strides(with_rank<N>, Layout, const Extents& ext, const Mapping& other)
 {
   static_assert(std::is_same<typename Mapping::layout_type, layout_stride>::value &&
-                (std::is_same<Layout, layout_left>::value or
+                (std::is_same<Layout, layout_left>::value ||
                  std::is_same<Layout, layout_right>::value)
                 , "This function is only intended to validate construction of "
                   "a layout_left or layout_right mapping from a layout_stride mapping.");
--- a/include/experimental/__p0009_bits/layout_stride.hpp
+++ b/include/experimental/__p0009_bits/layout_stride.hpp
@@ -605,7 +605,7 @@ struct layout_stride {
     )
     MDSPAN_INLINE_FUNCTION
     friend constexpr bool operator!=(const mapping& x, const StridedLayoutMapping& y) noexcept {
-      return not (x == y);
+      return !(x == y);
     }
 
     MDSPAN_TEMPLATE_REQUIRES(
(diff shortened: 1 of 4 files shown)


masterleinad:
Alternatively, you could do
```diff
diff --git a/include/experimental/__p0009_bits/macros.hpp b/include/experimental/__p0009_bits/macros.hpp
index 30209a6..34e2528 100644
--- a/include/experimental/__p0009_bits/macros.hpp
+++ b/include/experimental/__p0009_bits/macros.hpp
@@ -25,6 +25,10 @@
 #include "assert.h"
 #endif
 
+#ifdef _MSC_VER
+#include <iso646.h>
+#endif
+
 #ifndef _MDSPAN_HOST_DEVICE
 #  if defined(_MDSPAN_HAS_CUDA) || defined(_MDSPAN_HAS_HIP)
 #    define _MDSPAN_HOST_DEVICE __host__ __device__
```

masterleinad (approved):

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -643,7 +643,7 @@ constexpr void validate_strides(with_rank<0>, Layout, const Extents&, const Mapp
 template <std::size_t N, class Layout, class Extents, class Mapping>
 constexpr void validate_strides(with_rank<N>, Layout, const Extents& ext, const Mapping& other)
 {
-  static_assert(std::is_same<typename Mapping::layout_type, layout_stride>::value and
+  static_assert(std::is_same<typename Mapping::layout_type, layout_stride>::value &&
                 (std::is_same<Layout, layout_left>::value or
```
tpadioleau:
```suggestion
                (std::is_same<Layout, layout_left>::value ||
```

crtrott:
Let's avoid forcing default behavior change on downstream users.

tpadioleau (commented):
There is this clang-tidy check https://clang.llvm.org/extra/clang-tidy/checks/readability/operators-representation.html if this can help.

JBludau (approved):
all of them look correct to me

aprokop (approved):

tpadioleau (approved):

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -657,7 +657,7 @@ constexpr void validate_strides(with_rank<N>, Layout, const Extents& ext, const
     const std::size_t s = is_left ? r : N - 1 - r;
 
     MDSPAN_IMPL_PRECONDITION(common_integral_compare(expected_stride, other.stride(s))
-                             and "invalid strides for layout_{left,right}");
+                             && "invalid strides for layout_{left,right}");
```
tpadioleau, line 660: Just to make sure, do we intentionnally apply this binary AND between a boolean and a c-string ?

```cpp
MDSPAN_IMPL_PRECONDITION(common_integral_compare(expected_stride, other.stride(s)) && "invalid strides for layout_{left,right}")
```

mhoemmen:
I was going to say that including `<iso646.h>` is the suggested option.

https://learn.microsoft.com/en-us/cpp/c-runtime-library/iso646-operators?view=msvc-170

The alternative spellings are required by the C++ Standard.  Windows is nonconforming because it requires use of a header.  That being said, I don't object to the proposed solution.

crtrott:
Including that thing is transitive though, do we really want to change the default behavior of MSVC for everyone downstream?

include/experimental/__p0009_bits/layout_stride.hpp
crtrott: Yeah its a trick to get nice messages ...

mhoemmen:
> Including that thing is transitive though, do we really want to change the default behavior of MSVC for everyone downstream?

Just FYI, including that header doesn't "change the default behavior of MSVC," it makes MSVC conform to the C++ Standard a bit better.  That being said, I don't object to the proposed solution.

PR #348, opened 2024-06-18
CI: Enable warning flags

masterleinad:
It turns out that the builds in `cmake.yml` are mostly fine with `-Werror -Wall -Wextra -pedantic`:
- We get a bunch of warnings for unused variables building in `RelWithDebInfo` since that uses `-DNDEBUG` disabling a lot of asserts. Since we want these asserts to trigger, I changed the build type to `Debug`.
- `clang` complains about `-Wc++17-attribute-extensions` when using `[[maybe_unused]]` which seems harmless.
- `clang` complains about using variadic macros with empty ellipsis via `-Wno-gnu-zero-variadic-macro-arguments` which also seems harmless
- The other changes are about redundant semicolons or unused arguments.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -68,7 +68,7 @@ jobs:
     - name: Configure CMake
       shell: bash
       working-directory: ${{github.workspace}}/mdspan-build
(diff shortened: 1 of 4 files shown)


.github/workflows/cmake.yml
```diff
@@ -20,13 +20,13 @@ jobs:
...
-          cxx_flags_extra: "-stdlib=libc++"
+          cxx_flags_extra: "-stdlib=libc++ -Wno-c++17-attribute-extensions -Wno-gnu-zero-variadic-macro-arguments"
         - compiler_driver: icpx
           compiler_prefix: /opt/intel/oneapi/compiler/latest/linux/bin
           # To get new URL, look here:
           # https://www.intel.com/content/www/us/en/developer/articles/tool/oneapi-standalone-components.html#inpage-nav-6-undefined
           compiler_url: https://registrationcenter-download.intel.com/akdlm/IRC_NAS/ebf5d9aa-17a7-46a4-b5df-ace004227c0e/l_dpcpp-cpp-compiler_p_2023.2.1.8_offline.sh
-          cxx_flags_extra: "-DMDSPAN_USE_BRACKET_OPERATOR=0"
+          cxx_flags_extra: "-DMDSPAN_USE_BRACKET_OPERATOR=0 -Wno-c++17-attribute-extensions -Wno-gnu-zero-variadic-macro-arguments"
         - enable_benchmark: ON
```
nmm0, line 30: Won't switching to Debug affect the benchmark times? Maybe we should add build type to the matrix

masterleinad: Are these benchmark times recorded anywhere? To me, it seems we really only test correctness in the CI.

crtrott (approved):

PR #356, opened 2024-08-29
Disable empty return for _NVCOMPILER

tcclevenger:
These empty returns trigger `code_is_unreachable` warning for `nvc++` compiler. We want to incorporate changes in https://github.com/kokkos/kokkos/pull/7250.

```
"/var/jenkins/workspace/Kokkos_PR-7226/tpls/mdspan/include/mdspan/../experimental/__p2642_bits/layout_padded.hpp", line 64: warning: statement is unreachable [code_is_unreachable]
  return 0; 
  ^
          detected during:
            instantiation of class "Kokkos::Experimental::detail::static_array_type_for_padded_extent<_PaddingValue, _Extents, _ExtentToPadIdx, _Rank, Enabled> [with _PaddingValue=18446744073709551615UL, _Extents=Kokkos::extents<std::size_t, 18446744073709551615UL, 18446744073709551615UL>, _ExtentToPadIdx=0UL, _Rank=2UL, Enabled=void]" at line 95
            instantiation of class "Kokkos::Experimental::detail::padded_extent<_PaddingValue, _Extents, _ExtentToPadIdx> [with _PaddingValue=18446744073709551615UL, _Extents=Kokkos::extents<std::size_t, 18446744073709551615UL, 18446744073709551615UL>, _ExtentToPadIdx=0UL]" at line 174
            instantiation of class "Kokkos::Experimental::layout_left_padded<padding_value>::mapping<Extents> [with padding_value=18446744073709551615UL, Extents=Kokkos::extents<std::size_t, 18446744073709551615UL, 18446744073709551615UL>]" at line 883 of "/usr/include/c++/9/type_traits"
            instantiation of class "std::is_constructible<_Tp, _Args...> [with _Tp=Kokkos::Experimental::layout_left_padded<18446744073709551615UL>::mapping<Kokkos::extents<std::size_t, 18446744073709551615UL, 18446744073709551615UL>>, _Args=<Kokkos::extents<std::size_t, 18446744073709551615UL, 18446744073709551615UL>>]" at line 2912 of "/usr/include/c++/9/type_traits"
[...]

PR #358, opened 2024-09-03
MDSpan issues expose by Kokkos View refactor

crtrott:
- fixes and issue (and adds test) where `layout_left_paddded <-> layout_right_padded` conversion didn't work if the `extents_type` are not convertible
- uses correct constraint for pair like submdspan slices
- fixes a few missing things for CUDA and nvc++ compiler

Related to: https://github.com/kokkos/kokkos/pull/7288

### DIFF
--- a/include/experimental/__p2630_bits/submdspan_extents.hpp
+++ b/include/experimental/__p2630_bits/submdspan_extents.hpp
@@ -17,6 +17,7 @@
 #pragma once
 
 #include <tuple>
+#include <complex>
 
 #include "strided_slice.hpp"
 namespace MDSPAN_IMPL_STANDARD_NAMESPACE {
@@ -52,6 +53,33 @@ template <class OffsetType, class ExtentType, class StrideType>
 struct is_strided_slice<
     strided_slice<OffsetType, ExtentType, StrideType>> : std::true_type {};
 
+// Helper for identifying valid pair like things
+template <class T, class IndexType> struct index_pair_like : std::false_type {};
+
+template <class IdxT1, class IdxT2, class IndexType>
+struct index_pair_like<std::pair<IdxT1, IdxT2>, IndexType> {
+  static constexpr bool value = std::is_convertible_v<IdxT1, IndexType> &&
+                                std::is_convertible_v<IdxT2, IndexType>;
+};
+
+template <class IdxT1, class IdxT2, class IndexType>
+struct index_pair_like<std::tuple<IdxT1, IdxT2>, IndexType> {
+  static constexpr bool value = std::is_convertible_v<IdxT1, IndexType> &&
+                                std::is_convertible_v<IdxT2, IndexType>;
+};
+
+template <class IdxT, class IndexType>
+struct index_pair_like<std::complex<IdxT>, IndexType> {
+  static constexpr bool value = std::is_convertible_v<IdxT, IndexType>;
+};
+
+template <class IdxT, class IndexType>
+struct index_pair_like<std::array<IdxT, 2>, IndexType> {
+  static constexpr bool value = std::is_convertible_v<IdxT, IndexType>;
+};
+
+// FIXME: we actually need to pass IndexType into all of these
+
 // first_of(slice): getting begin of slice specifier range
 MDSPAN_TEMPLATE_REQUIRES(
   class Integral,
@@ -70,7 +98,7 @@ first_of(const ::MDSPAN_IMPL_STANDARD_NAMESPACE::full_extent_t &) {
 
 MDSPAN_TEMPLATE_REQUIRES(
   class Slice,
-  /* requires */(std::is_convertible_v<Slice, std::tuple<size_t, size_t>>)
+  /* requires */(index_pair_like<Slice, size_t>::value)
 )
 MDSPAN_INLINE_FUNCTION
 constexpr auto first_of(const Slice &i) {
@@ -100,7 +128,7 @@ constexpr Integral
 
 MDSPAN_TEMPLATE_REQUIRES(
   size_t k, class Extents, class Slice,
-  /* requires */(std::is_convertible_v<Slice, std::tuple<size_t, size_t>>)
+  /* requires */(index_pair_like<Slice, size_t>::value)
 )
 MDSPAN_INLINE_FUNCTION
 constexpr auto last_of(std::integral_constant<size_t, k>, const Extents &,
--- a/include/experimental/__p2630_bits/submdspan_extents.hpp
+++ b/include/experimental/__p2630_bits/submdspan_extents.hpp
@@ -78,8 +78,6 @@ struct index_pair_like<std::array<IdxT, 2>, IndexType> {
   static constexpr bool value = std::is_convertible_v<IdxT, IndexType>;
 };
 
-// FIXME: we actually need to pass IndexType into all of these
-
 // first_of(slice): getting begin of slice specifier range
 MDSPAN_TEMPLATE_REQUIRES(
   class Integral,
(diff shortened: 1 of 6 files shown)


nmm0 (commented):
This includes changes from https://github.com/kokkos/kokkos/pull/7250 right?

include/experimental/__p2642_bits/layout_padded.hpp
```diff
@@ -95,6 +95,7 @@ struct padded_extent {
   using static_array_type = typename static_array_type_for_padded_extent<
       padding_value, _Extents, _ExtentToPadIdx, _Extents::rank()>::type;
 
+  MDSPAN_INLINE_FUNCTION
```
masterleinad, line 98: Do you actually need this or did you just add for consistency?

include/experimental/__p2642_bits/layout_padded.hpp
```diff
@@ -136,7 +137,7 @@ struct padded_extent {
       return {};
     }
     // Missing return statement warning from NVCC and ICC
-#if defined(__NVCC__) || defined(__INTEL_COMPILER)
+#if (defined(__NVCC__) || defined(__INTEL_COMPILER)) && !defined(__NVCOMPILER)
     return {};
 #endif
```
masterleinad: All of these are contained in https://github.com/kokkos/mdspan/pull/356 and should disappear after rebasing.

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -52,6 +53,33 @@ template <class OffsetType, class ExtentType, class StrideType>
...
+template <class IdxT1, class IdxT2, class IndexType>
+struct index_pair_like<std::tuple<IdxT1, IdxT2>, IndexType> {
+  static constexpr bool value = std::is_convertible_v<IdxT1, IndexType> &&
+                                std::is_convertible_v<IdxT2, IndexType>;
+};
+
+template <class IdxT, class IndexType>
+struct index_pair_like<std::complex<IdxT>, IndexType> {
+  static constexpr bool value = std::is_convertible_v<IdxT, IndexType>;
+};
```
masterleinad, line 74: `std::get<0>(z)` is only available from C++26 on, right? https://en.cppreference.com/w/cpp/numeric/complex/get

include/experimental/__p2642_bits/layout_padded.hpp
```diff
@@ -203,7 +204,7 @@ class layout_left_padded<PaddingValue>::mapping {
   }
 
 public:
-#if !MDSPAN_HAS_CXX_20
+#if !MDSPAN_HAS_CXX_20 || defined(__NVCC__)
```
masterleinad, line 207: What's the problem here?

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -52,6 +53,33 @@ template <class OffsetType, class ExtentType, class StrideType>
 struct is_strided_slice<
     strided_slice<OffsetType, ExtentType, StrideType>> : std::true_type {};
 
+// Helper for identifying valid pair like things
+template <class T, class IndexType> struct index_pair_like : std::false_type {};
```
masterleinad, line 57: The feature macro for complex `get` support is called [__cpp_lib_tuple_like](https://en.cppreference.com/w/cpp/feature_test#cpp_lib_tuple_like) so maybe
```suggestion
template <class T, class IndexType> struct index_tuple_like : std::false_type {};
```

include/experimental/__p2630_bits/submdspan_extents.hpp
crtrott: It also wouldn't work portably. In fact we need to overhaul more stuff because `integral_constant::value()` and `get<>(tuple)` give warnings.

crtrott: And just to be clear: the mdspan specification uses `pair-like`,  and that is a named list of types, not everything which supports tuple protocol and has size 2. Hence, my change in the implementation here.

include/experimental/__p2642_bits/layout_padded.hpp
crtrott: `static_value` is called inside a couple constructors and `is_always_exhaustive` all of which are device callable. The "INLINE" is just for consistency, at some point we can go wholesale through this stuff and replace it all with MDSPAN_FUNCTION

include/experimental/__p2630_bits/submdspan_extents.hpp
crtrott: this has nothing to do with get: https://eel.is/c++draft/views.multidim#mdspan.syn
We are using an exposition only concept called `index-pair-like` and if you follow all the stuff down to their roots, this thing here basically implements that concept in form of a `type_trait`.

include/experimental/__p2642_bits/layout_padded.hpp
crtrott: NVCC does not like the requires stuff down there. Basically, it ends up saying that there is not default constructor on the device available in C++20 mode.

crtrott: I wasn't able to narrow it down in a simple reproducer.

include/experimental/__p2642_bits/layout_padded.hpp
crtrott: rebased

include/experimental/__p2630_bits/submdspan_extents.hpp
masterleinad: The reason why I looked at `get` was that that's what is used in `first_of` which now uses this concept as requirement.

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -52,6 +53,33 @@ template <class OffsetType, class ExtentType, class StrideType>
...
+struct index_pair_like<std::complex<IdxT>, IndexType> {
+  static constexpr bool value = std::is_convertible_v<IdxT, IndexType>;
+};
+
+template <class IdxT, class IndexType>
+struct index_pair_like<std::array<IdxT, 2>, IndexType> {
+  static constexpr bool value = std::is_convertible_v<IdxT, IndexType>;
+};
+
+// FIXME: we actually need to pass IndexType into all of these
```
dalg24: What is left to do?

include/experimental/__p2630_bits/submdspan_extents.hpp
crtrott: I fixed that by adding special versions for complex.

masterleinad (approved):
Fine with me.

include/experimental/__p2630_bits/submdspan_extents.hpp
dalg24: It is not because it will happen to work in C++26 for complex that we should get out of our way, pull in the header for `<complex>` and specialize.
This is not necessary for the Kokkos refactor.  Please withdraw the header include and that specialization and propose it as a follow up if you really want it.

dalg24 (changes requested):

include/experimental/__p2630_bits/submdspan_extents.hpp
masterleinad: At least `std::get` isn't used with `complex` anymore after the last commits.

crtrott:
As discussed elsewhere: complex is one of the named types for which this is supposed to work based on the definition of index-pair-like I see no reason to not include it here in the back port. Every standard approved version of submdspan will take complex as slice argument, and so I don't see a good reason to exclude it from this back port to C++17.

dalg24 (approved):

PR #360, opened 2024-10-02
Fix device/host call issues in CUDA (when not using relaxed constexpr workaround)

nmm0:
This includes code that Christian wrote and will be included into Kokkos in https://github.com/kokkos/kokkos/pull/7396

This addresses warnings for calling constexpr device code inside of Kokkos with BasicView


include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -199,7 +199,15 @@ struct layout_stride {
       template<class IntegralType>
       MDSPAN_INLINE_FUNCTION
       static constexpr const __strides_storage_t fill_strides(const std::array<IntegralType,extents_type::rank()>& s) {
+        // avoid warning for use of host std::array operator[]
+  #if defined(_MDSPAN_HAS_CUDA) || defined(_MDSPAN_HAS_HIP)
+        const IntegralType* s_ptr = reinterpret_cast<const IntegralType*>(&s);
```
nmm0: We need to figure out what to do properly here, since `reinterpret_cast` can't be used in a constant expression, so we will remove it for now

nmm0:
- [x] Move `device_support` to `utility`
- [x] Make `construct_sub_strides` use a custom c-array like type instead of `std::array`
- [x] Remove fake-tuple type access
- [x] Make std::array constructors for layout_stride non-device

include/experimental/__p0009_bits/layout_stride.hpp
masterleinad: `std::array` isn't usable on Cuda devices anyway unless you are using `-expt-relaxed-constexpr`, right? Wouldn't an interface in terms of `C-style` arrays be a more natural choice?

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -206,7 +205,7 @@ struct layout_stride {
         class IntegralType,
         // The is_convertible condition is added to make sfinae valid
         // the extents_type::rank() > 0 is added to avoid use of non-standard zero length c-array
```
tcclevenger:
```suggestion
```
This part of the comment is no longer needed.

include/experimental/__p0009_bits/layout_stride.hpp
crtrott: Yes and no: we have constructors which have to take std::array because that is the standard. Also a bunch of operations actually do work on device (i.e. construction, destruction, structured binding - because they don't go through member functions, or at least go through defaulted member functions).

That said the above casting approach doesn't work in constexpr so we changed this.

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -206,7 +205,7 @@ struct layout_stride {
...
-        (std::is_convertible<IntegralType, typename extents_type::index_type>::value && (extents_type::rank() > 0))
+        (std::is_convertible<IntegralType, typename extents_type::index_type>::value)
       )
       MDSPAN_INLINE_FUNCTION
       // despite the requirement some compilers still complain about zero length array during parsing
```
masterleinad: This comment seems out-of-date now.

include/experimental/__p0009_bits/layout_stride.hpp
masterleinad:
```suggestion
```

include/experimental/__p0009_bits/utility.hpp
```diff
@@ -64,6 +66,112 @@ constexpr struct
   }
 } stride;
 
+template<class T>
+MDSPAN_INLINE_FUNCTION
+constexpr void maybe_unused_variable(const T&) {}
```
masterleinad: This function seems unused.
```suggestion
```

tcclevenger (commented):
Just curious, if C++17 is the minimum req for Kokkos core, why do we test C++14 for mdspan. Is this the same for other tpls or containers/algorithms?

include/experimental/__p0009_bits/utility.hpp
```diff
@@ -64,6 +66,112 @@ constexpr struct
...
+template <class T, T v>
+struct integral_constant {
+  using value_type         = T;
+  using type               = integral_constant<T, v>;
+
+  MDSPAN_INLINE_FUNCTION_DEFAULTED
+  constexpr integral_constant() = default;
+  
+  MDSPAN_INLINE_FUNCTION_DEFAULTED
+  constexpr integral_constant(std::integral_constant<T,v>) {};
```
masterleinad: `DEFAULTED` seems inappropriate. Do you need the extra annotations at all here?
```suggestion
  MDSPAN_INLINE_FUNCTION
  constexpr integral_constant(std::integral_constant<T,v>) {};
```

include/experimental/__p0009_bits/layout_stride.hpp
crtrott: There are users of mdspan which are not Kokkos users and rely on C++14.

crtrott: but yeah good observation I just pushed something to remove the comments

include/experimental/__p0009_bits/utility.hpp
```diff
@@ -64,6 +66,112 @@ constexpr struct
...
+  MDSPAN_FUNCTION constexpr T& get() { return val; }
+  MDSPAN_FUNCTION constexpr const T& get() const { return val; }
+};
+
+template<size_t SearchIdx, size_t Idx, class T>
+struct tuple_idx_matcher {
+  using type = tuple_member<T, Idx>;
+  template<class Other>
+  MDSPAN_FUNCTION
+  constexpr auto operator + (Other v) const {
```
masterleinad: Using `+` seems a little weird. What `|` or `||` be slightly better?

include/experimental/__p0009_bits/utility.hpp
```diff
@@ -64,6 +66,112 @@ constexpr struct
...
+  using type               = integral_constant<T, v>;
+
+  MDSPAN_INLINE_FUNCTION_DEFAULTED
+  constexpr integral_constant() = default;
+  
+  MDSPAN_INLINE_FUNCTION_DEFAULTED
+  constexpr integral_constant(std::integral_constant<T,v>) {};
+
+  static constexpr T value = v;
+  MDSPAN_INLINE_FUNCTION constexpr operator value_type() const noexcept {
```
dalg24: Prefer `MDSPAN_FUNCTION` here and below

include/experimental/__p0009_bits/utility.hpp
```diff
@@ -64,6 +66,112 @@ constexpr struct
   }
 } stride;
 
+template<class T>
+MDSPAN_INLINE_FUNCTION
+constexpr void maybe_unused_variable(const T&) {}
+
+
+// same as std::integral_constant but with __host__ __device__ annotations on
+// the implicit conversion function and the call operator
```
dalg24, line 70: Why do we need/want conversion from to the std:: counterpart?

include/experimental/__p0009_bits/utility.hpp
```diff
@@ -64,6 +66,112 @@ constexpr struct
...
+  constexpr integral_constant(std::integral_constant<T,v>) {};
+
+  static constexpr T value = v;
+  MDSPAN_INLINE_FUNCTION constexpr operator value_type() const noexcept {
+    return value;
+  }
+  MDSPAN_INLINE_FUNCTION constexpr value_type operator()() const noexcept {
+    return value;
+  }
+  MDSPAN_INLINE_FUNCTION constexpr operator std::integral_constant<T,v>() const noexcept {
```
dalg24: No way that can be annotated host device

include/experimental/__p0009_bits/utility.hpp
dalg24: Can't be host device in the 1st place

crtrott: yes it can. Nothing in this calls device functions. Constructor of std::integral_constant is defaulted, so it works too. The only thing which doesn't work of std::integral_constant is the conversion operator to its scaler type.

include/experimental/__p0009_bits/utility.hpp
crtrott: we just need something which expands.

include/experimental/__p0009_bits/utility.hpp
crtrott: because there are places where we pass in std::integral_constant, and std::integral_constant is the standard approved way to pass certain ways to certain functions. For example `submdspan(a, std::integral_constant<int, 1>())` is fine in device code.

include/experimental/__p0009_bits/utility.hpp
crtrott: Of course it works: 

https://godbolt.org/z/bb8qdT31Y

We can use any standard library stuff as long as we don't call non-defaulted member functions. And we don't here.

include/experimental/__p0009_bits/utility.hpp
dalg24: Then what Daniel said

dalg24: (minus the inline)

include/experimental/__p0009_bits/utility.hpp
```diff
@@ -64,6 +66,112 @@ constexpr struct
...
+template<size_t Idx, class ... Args>
+MDSPAN_FUNCTION
+constexpr const auto& get(const tuple<Args...>& vals) { return vals.template get<Idx>(); }
+
+template<class ... Elements>
+tuple(Elements ...) -> tuple<Elements...>;
+#endif
+
+template<class T, size_t ... Idx>
+constexpr auto c_array_to_std(std::index_sequence<Idx...>, const T(&values)[sizeof...(Idx)]) {
```
dalg24: I would prefer calling it `to_std_array`

dalg24: Also thinking you could should write a for loop instead of a fold expression

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -84,14 +85,27 @@ any_slice_out_of_bounds(const extents<IndexType, Exts...> &exts,
 }
 
 // constructs sub strides
+template<class T, size_t N>
+struct sub_strides
+{
+  T values[N];
```
dalg24:
```suggestion
  T values[N > 0 ? N : 1];
```

crtrott: Done (and delete the specialization below) unless you prefer the specialization path.

include/experimental/__p0009_bits/utility.hpp
crtrott: got rid of the function for now, not really needed for what we do.

include/experimental/__p0009_bits/utility.hpp
crtrott: done

include/experimental/__p0009_bits/utility.hpp
crtrott: yeah clear bug (funnily enough one of the primary bugs we had elsewhere)

include/experimental/__p0009_bits/utility.hpp
crtrott: removed

include/experimental/__p0009_bits/layout_stride.hpp
crtrott: removed

include/experimental/__p0009_bits/layout_stride.hpp
crtrott: removed

masterleinad (approved):

crtrott (approved):

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -476,7 +468,8 @@ struct layout_stride {
     MDSPAN_INLINE_FUNCTION
     constexpr index_type required_span_size() const noexcept {
       index_type span_size = 1;
-      for(unsigned r = 0; r < extents_type::rank(); r++) {
+      // using int here to avoid warning about pointless comparison to 0
```
dalg24, line 471: I suppose you mean the rank 0 case

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -68,6 +69,12 @@ struct index_pair_like<std::tuple<IdxT1, IdxT2>, IndexType> {
                                 std::is_convertible_v<IdxT2, IndexType>;
 };
 
+template <class IdxT1, class IdxT2, class IndexType>
+struct index_pair_like<tuple<IdxT1, IdxT2>, IndexType> {
```
dalg24, line 73: Might be worth a comment "our device-compatible poor man's implementation of tuple" for the future reader that might be surprised to see that specialization interleaved with all the std:: ones.

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -88,10 +95,16 @@ constexpr Integral first_of(const Integral &i) {
   return i;
 }
 
+template<class Integral, Integral v>
+MDSPAN_INLINE_FUNCTION
+constexpr Integral first_of(const std::integral_constant<Integral, v>&) {
+  return integral_constant<Integral, v>();
+}
```
dalg24, line 102: This is weird

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -100,7 +113,24 @@ MDSPAN_TEMPLATE_REQUIRES(
...
+  return get<0>(i);
+}
+
+MDSPAN_TEMPLATE_REQUIRES(
+  class IdxT1, class IdxT2,
+  /* requires */ (index_pair_like<std::pair<IdxT1, IdxT2>, size_t>::value)
+  )
+MDSPAN_INLINE_FUNCTION
+constexpr auto first_of(const std::pair<IdxT1, IdxT2>& i) {
+  return i.first;
```
dalg24, line 133: Can't we use get<I>() for all of these

include/experimental/__p2630_bits/submdspan_extents.hpp
dalg24: Why do we need this one

include/experimental/__p2630_bits/submdspan_extents.hpp
crtrott: get on std::pair is not host-device capable.

PR #366, opened 2024-11-26
unstdify config.hpp macros

nmm0:
I used the [bugprone-reserved-identifier](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/reserved-identifier.html) clang-tidy check to find these, and a global search/replace to fix them. I kept these changes isolated to ones _declared_ in config.hpp and plan on doing more later. (As you can see, even just the config.hpp ones touch basically every file)

### DIFF
--- a/include/experimental/__p0009_bits/config.hpp
+++ b/include/experimental/__p0009_bits/config.hpp
@@ -31,9 +31,9 @@
 #endif
 
 #ifdef _MSVC_LANG
-#define MDSPAN_CPLUSPLUS _MSVC_LANG
+#define MDSPAN_IMPL_CPLUSPLUS _MSVC_LANG
 #else
-#define MDSPAN_CPLUSPLUS __cplusplus
+#define MDSPAN_IMPL_CPLUSPLUS __cplusplus
 #endif
 
 #define MDSPAN_CXX_STD_14 201402L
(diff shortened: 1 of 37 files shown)


include/experimental/__p0009_bits/config.hpp
```diff
@@ -15,21 +15,25 @@
 //@HEADER
 #pragma once
 
-#ifndef __has_include
-#  define __has_include(x) 0
+#ifndef MDSPAN_IMPL_HAS_INCLUDE
```
nmm0, line 18: I switched this to our own macro rather than substituting in the standard one (since I think that could propagate a bug downstream if someone does `#ifdef __has_include`

include/experimental/__p0009_bits/config.hpp
```diff
@@ -42,190 +46,194 @@
...
+#ifndef MDSPAN_HAS_SYCL
 #  if defined(SYCL_LANGUAGE_VERSION)
-#    define _MDSPAN_HAS_SYCL SYCL_LANGUAGE_VERSION
+#    define MDSPAN_HAS_SYCL SYCL_LANGUAGE_VERSION
 #  endif
 #endif
 
-#ifndef __has_cpp_attribute
-#  define __has_cpp_attribute(x) 0
+#ifndef MDSPAN_IMPL_HAS_CPP_ATTRIBUTE
```
nmm0, line 102: Same here

dalg24 (commented):
We shouldn't just remove the leading underscore, we need to inject something to denote that these macros are private, along the lines of what we do in Kokkos with the KOKKOS_IMPL_ prefix

include/experimental/__p0009_bits/config.hpp
mhoemmen: This is much better; thank you! : - )

mhoemmen (approved):

nmm0:
> We shouldn't just remove the leading underscore, we need to inject something to denote that these macros are private, along the lines of what we do in Kokkos with the KOKKOS_IMPL_ prefix

@dalg24 ok, I can do `MDSPAN_IMPL_`

dalg24 (approved):

PR #377, opened 2025-01-16
GitHub CI: pin ubuntu-22.04

masterleinad:


nmm0:
I guess we need to pin the OS version as `ubuntu-latest` is a moving target. What if we used some of the kokkos images?

masterleinad:
> I guess we need to pin the OS version as `ubuntu-latest` is a moving target. What if we used some of the kokkos images?

Looks like just pinning ubuntu-22.04 is sufficient for now and we could drop forcing`-stdlib=libc++` and linking to `libbenchmark1` (but I reverted those changes again).

nmm0 (approved):

PR #379, opened 2025-02-14
remove cmake logic for `-fconcepts` and `-fconcepts-ts` flags

nmm0:
I'm suggesting that we remove the logic in our cmake that tries to add the `-fconcepts` and `-fconcepts-ts` flags.

First of all, the `-fconcepts-ts` branch is broken, because the condition checks `COMPILER_SUPPORTS_FCONCEPTS` instead of `COMPILER_SUPPORTS_FCONCEPTS_TS`, so it wasn't used anyway.

Secondly, pretty much every modern compiler will enable this if you enable `-std=c++20`, and we only add this flag if we enable `-std=c++20`. In the rare cases someone is trying to build mdspan on a platform that doesn't do this, they can simply add `-DCMAKE_CXX_FLAGS=-fconcepts` to their cmake command line invocation or preset.

Note that us manually adding this flag breaks builds in small niche use cases (that can easily be worked around); i.e. when using `CMAKE_CXX_CLANG_TIDY` with gcc as clang does not recognize the flag but it is forced on. So the motivation for this is pretty minor but I think it would be good to clean up some of our cmake anyway.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,27 +81,6 @@ endif()
 
 ################################################################################
 
-if(MDSPAN_ENABLE_CONCEPTS)
-  if(CMAKE_CXX_STANDARD GREATER_EQUAL 20)
-    include(CheckCXXCompilerFlag)
-    check_cxx_compiler_flag("-fconcepts" COMPILER_SUPPORTS_FCONCEPTS)
-    if(COMPILER_SUPPORTS_FCONCEPTS)
-      message(STATUS "Using \"-fconcepts\" to enable concepts support")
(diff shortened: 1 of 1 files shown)


nmm0:
Also, GCC 15 [is removing](https://gcc.gnu.org/gcc-15/changes.html) support for `-fconcepts-ts`. It won't have any effect anymore.

mhoemmen (approved):
Thanks @nmm0 ! : - )  [config.hpp](https://github.com/kokkos/mdspan/blob/stable/include/experimental/__p0009_bits/config.hpp) doesn't seem to use this CMake option anyway, so removing this shouldn't change the code.

dalg24 (approved):
Thanks for cleaning that up

PR #380, opened 2025-02-14
Use fetchcontent googletest

nmm0:
FetchContent was introduced in cmake 3.11, we require 3.14 and above. FetchContent is pretty much a drop-in replacement for our previous code.

I also switched the target we use to the namespaced target: `GTest::gtest_main`. This is to avoid potential issues with accidentally adding `-lgtest_main` to the command line and finding a system gtest_main in some situations.


mhoemmen (approved):
EXCELLENT

PR #381, opened 2025-02-14
Add fetchcontent for google benchmark

nmm0:
This provides a way of getting google benchmark via FetchContent.

I added a flag `MDSPAN_USE_SYSTEM_BENCHMARK` to control whether to use a system-provided library or not. As the previous behavior was to use a system library, I defaulted this to `On` (which is the opposite of what we do for googletest). I'm open to changing the default to `Off` if we think that is more desirable.


mhoemmen (approved):
Thanks for contributing! : - )

mhoemmen (approved):

PR #388, opened 2025-03-17
Unstdify all the things

nmm0:
This makes clang-tidy run clean with the following parameters, and is (hopefully) the last part of my cleanup PRs:

```sh
clang-tidy -p cmake-build-debug --checks="-*,bugprone-reserved-identifier" -warnings-as-errors="*" --header-filter="include/*" include/mdspan/*.hpp
```

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,4 @@
+Checks: >
+  -*,
+  bugprone-reserved-identifier
+HeaderFilterRegex: 'include/*'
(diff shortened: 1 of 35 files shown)


masterleinad (commented):
I would actually add the `.clang-tidy` file with the checks you enabled.
Also compilation seeems to fail with `C++14` support.

nmm0:
> I would actually add the `.clang-tidy` file with the checks you enabled. Also compilation seeems to fail with `C++14` support.

I was thinking of doing it in a separate PR after this is merged just to make it less large.

Yeah, this PR kinda exploded in CI which was weird because I hadn't been seeing any issues on my local machine. I tested with 17, 20, and 23 but not 14 though still some of the gcc 9 and 10 builds are failing.

crtrott (approved):

masterleinad (approved):

PR #391, opened 2025-03-19
fix incorrect `required_span_size` for padded layouts

### LINKED ISSUE #362: Possible bug in `required_span_size`
I've come across an issue which I don't completely understand but which feels like a bug in `required_span_size`. A working example is here: https://godbolt.org/z/Y8bcnGeEj

My understanding of `required_span_size` is that it allows a span to be constructed which contains all the referenceable data from an `mdspan`. So if I have an `mdspan` called `m`, its elements will all be contained in `std::span(m.data_handle(), required_span_size())`  and that the span will be minimal (i.e., just big enough to store the data without padding on the end). Is that the correct intention?

Here is a cut-down reproducer for the issue I see:

```cpp
// Original mdspan across a 4x5 data block.
std::mdspan<float, std::extents<size_t, 4, 5>> m1(array);

// The bottom-right 2x2 matrix
auto m2 = submdspan(m1, std::pair(2, 4), std::pair(3, 5));
    
// Create a span from the submdspan. BUT, this span goes past the end of  `array` 
auto m2span = std::span(m2.data_handle(), m2.mapping().required_span_size());
// required_span_size() is bigger than m2.mapping()(1, 1) + 1
```

The final span is outside the bounds of the original parent data so it allows access to data it shouldn't be able to get at.

nmm0:
Fixes #362

### DIFF
--- a/include/experimental/__p2642_bits/layout_padded.hpp
+++ b/include/experimental/__p2642_bits/layout_padded.hpp
@@ -390,7 +390,7 @@ public:
       for (rank_type r = 1; r < extents_type::rank(); ++r) {
         value *= exts.extent(r);
       }
-      return value;
+      return value + exts.extent(0) - padded_stride.value(0);
     }
   }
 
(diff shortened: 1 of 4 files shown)


mhoemmen (approved):
Thanks for the fix! : - )

PR #395, opened 2025-03-31
fix padded layouts for non-size_t indices

### LINKED ISSUE #393: Add precondition checks and mandates to test padding representability
We are not dealing with convertability properly for non-size_t indices in padded layouts and not enforcing mandates on the representability of PaddingValue:

http://eel.is/c++draft/mdspan.layout#leftpad.overview-5

Basically we need to enforce these constraints or we could potentially get subtle overflow bugs if a user is providing a smaller index_type (like char or something)

nmm0:
Fixes #393


tests/test_layout_padded_left.cpp
```diff
@@ -1,9 +1,19 @@
...
+
+namespace {
+static bool precondition_failed = false;
+void handle_test_precondition_failure(const char *, const char *, unsigned) {
+  precondition_failed = true;
+}
+} // namespace
+
+#define MDSPAN_IMPL_PRECONDITION_VIOLATION_HANDLER(cond, file, line) \
+  handle_test_precondition_failure(cond, file, line)
```
nmm0: Is this construct worth making more common to other tests?

crtrott (changes requested):
For precondition checks all of these things need to be device functions.

include/experimental/__p0009_bits/utility.hpp
```diff
@@ -165,6 +165,63 @@ constexpr const auto& get(const tuple<Args...>& vals) { return vals.template get
...
+    // check overflow for negative a, b
+    if (b < 0 && a < std::numeric_limits<T>::max() / b)
+      return false;
+    // check underflow for positive b, negative a
+    if (b > 0 && a < std::numeric_limits<T>::min() / b)
+      return false;
+    // check underflow for negative b, positive a
+    if (b < 0 && a > std::numeric_limits<T>::min() / b)
+      return false;
+  }
```
crtrott: This is a function which is only used with extents right now. Let's not pay for the overhead of all these other checks for negative number combos.

```suggestion
  if constexpr (std::is_signed_v<T>) {
    if ( a < 0 || b < 0 ) return false;
  }
  return a <= std::numeric_limits<T>::max() / b;
```

crtrott (changes requested):
I think we can make the check cheaper since we only use it for extents. Otherwise looks good. I made a suggestion.

include/experimental/__p0009_bits/utility.hpp
crtrott: possibly could rename function to indicate we expect positive.

crtrott (approved):

PR #397, opened 2025-04-04
Guard SYCL printf extension

### LINKED ISSUE #396: `sycl::ext::oneapi::experimental::printf` is an intel extension
Hi,

https://github.com/kokkos/mdspan/blob/ae4d3ff8268ee7159f1e85f53b2db00e05ad5493/include/experimental/__p0009_bits/macros.hpp#L121

`sycl::ext::oneapi::experimental::printf` is an Intel SYCL extension and is not supported by other implementations (for example, https://github.com/AdaptiveCpp/AdaptiveCpp). 

I sugest guarding it with `__INTEL_LLVM_COMPILER` (https://www.intel.com/content/www/us/en/developer/articles/technical/use-predefined-macros-for-specific-code-for-intel-dpcpp-compiler-intel-cpp-compiler-intel-cpp-compiler-classic.html) 

Can make the PR if you agree with the problem and fix

TApplencourt:
Fix https://github.com/kokkos/mdspan/issues/396

### DIFF
--- a/include/experimental/__p0009_bits/macros.hpp
+++ b/include/experimental/__p0009_bits/macros.hpp
@@ -120,6 +120,10 @@ MDSPAN_FUNCTION inline void default_precondition_violation_handler(const char* c
 {
 #ifdef __INTEL_LLVM_COMPILER
   sycl::ext::oneapi::experimental::printf("%s:%u: precondition failure: `%s`\n", file, line, cond);
+#else
+  (void) cond;
+  (void) file;
+  (void) line;
 #endif
   assert(0);
 }
--- a/include/experimental/__p0009_bits/macros.hpp
+++ b/include/experimental/__p0009_bits/macros.hpp
@@ -118,7 +118,9 @@ MDSPAN_FUNCTION inline void default_precondition_violation_handler(const char* c
 #elif defined(MDSPAN_IMPL_HAS_SYCL)
 MDSPAN_FUNCTION inline void default_precondition_violation_handler(const char* cond, const char* file, unsigned line)
 {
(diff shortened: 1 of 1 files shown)


masterleinad (approved):

include/experimental/__p0009_bits/macros.hpp
```diff
@@ -118,7 +118,9 @@ MDSPAN_FUNCTION inline void default_precondition_violation_handler(const char* c
 #elif defined(MDSPAN_IMPL_HAS_SYCL)
 MDSPAN_FUNCTION inline void default_precondition_violation_handler(const char* cond, const char* file, unsigned line)
```
masterleinad, line 119: All arguments are unused if the compiler isn't `IntelLLVM` which might result in compiler warnings.

TApplencourt: Good point.
Do you prefer me to `(void) foo` them or to use some attribute to disable warning for this function / argument (GCC `__attribute__((unused))`) or fancy C++17 attribute `[[maybe_unused]]` ?

masterleinad: Looks like this file uses `(void) bla` instead of `[[maybe_unused]]`.

TApplencourt: Sorry, I'm stupid. I should have just looked too.

TApplencourt: Fixed with b86c4f1c6955e2ff6bdd3abec048dc1c846527a1

dalg24 (approved):

PR #400, opened 2025-04-16
Add mdspan at (rebase + fixup)

### LINKED ISSUE #300: Access with bounds checking using `::at`

nmm0:
This is a rebase and fixup of https://github.com/kokkos/mdspan/pull/302 which implements [P3383](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p3383r2.html)

Thanks to the authors, including @stephanlachnit who wrote the original PR.

Fixes #300

### DIFF
--- a/include/experimental/__p0009_bits/config.hpp
+++ b/include/experimental/__p0009_bits/config.hpp
@@ -184,6 +184,13 @@ static_assert(MDSPAN_IMPL_CPLUSPLUS >= MDSPAN_CXX_STD_14, "mdspan requires C++14
 #  endif
 #endif
 
+#ifndef _MDSPAN_USE_IF_CONSTEXPR_17
+#  if (defined(__cpp_if_constexpr) && __cpp_if_constexpr >= 201606) \
+        || (!defined(__cpp_constexpr) && MDSPAN_HAS_CXX_17)
+#    define _MDSPAN_USE_IF_CONSTEXPR_17 1
+#  endif
+#endif
+
 #ifndef MDSPAN_IMPL_USE_INTEGER_SEQUENCE_14
 #  if defined(MDSPAN_IMPL_COMPILER_MSVC)
 #    if (defined(__cpp_lib_integer_sequence) && __cpp_lib_integer_sequence >= 201304)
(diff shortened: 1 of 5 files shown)


include/experimental/__p0009_bits/config.hpp
```diff
@@ -184,6 +184,13 @@ static_assert(MDSPAN_IMPL_CPLUSPLUS >= MDSPAN_CXX_STD_14, "mdspan requires C++14
 #  endif
 #endif
 
+#ifndef MDSPAN_IMPL_USE_IF_CONSTEXPR_17
```
mhoemmen, line 187: Do we actually test mdspan with C++14 any more?  If not, then there's no need to have a macro for this feature, and the macro wouldn't be tested anyway.

mhoemmen (approved):
Thanks for contributing this fix! : - )

My only question is whether we have testing for C++14.  If not, then there's no reason to have a back-porting macro for `if constexpr` support.

include/experimental/__p0009_bits/config.hpp
nmm0: Yes we still have some CI builds (there is at least one user who still uses C++14). We don't guarantee that everything still works with C++14 though and guard against that. For example padded layouts will not work with 14. The MDSpan class does need to so unfortunately we still need this switch

mhoemmen: Thanks for explaining!  The macro is an implementation detail so it's fine.

crtrott (approved):

PR #403, opened 2025-05-12
Avoid static variables and functions in header files

masterleinad:
Extracted from https://github.com/kokkos/kokkos/pull/8071. As discussed there (and in https://github.com/kokkos/kokkos/pull/7898) `static` global variables and functions are problematic when building C++20 modules.

### DIFF
--- a/include/experimental/__p0009_bits/utility.hpp
+++ b/include/experimental/__p0009_bits/utility.hpp
@@ -47,7 +47,10 @@ constexpr bool rankwise_equal(with_rank<N>, const T1& x, const T2& y, F func)
   return match;
 }
 
-inline constexpr struct
+#if MDSPAN_HAS_CXX_17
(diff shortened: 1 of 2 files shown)


include/experimental/__p0009_bits/utility.hpp
```diff
@@ -167,6 +173,9 @@ tuple(Elements ...) -> tuple<Elements...>;
 #endif
 } // namespace detail
 
+#if MDSPAN_HAS_CXX_17
+inline
```
nmm0, line 177: Do these not have multiple definition problems without the inline?

masterleinad: A `const` variable at namespace scope has internal linkage by default.

dalg24 (approved):

PR #408, opened 2025-05-20
Implement P3663 (Future-proof submdspan_mapping)

mhoemmen:
Implement P3663 (Future-proof `submdspan_mapping`).

- [X] Fix merge conflicts
- [X] Fix all check-in builds when `MDSPAN_ENABLE_P3663=OFF` (default)
- [x] Fix all check-in builds when `MDSPAN_ENABLE_P3663=ON` (this may require significant back-porting effort)
    - [x] There's probably no way to get anything resembling `constant_wrapper` in C++17 or less.  Even the C++20 "back-port" is so incomplete that it might as well be called "`integral_constant`."  We could either just use `integral_constant` for C++20 or less, or name the type `detail::constant_wrapper`.
- [x] Make `MDSPAN_ENABLE_P3663=ON` the default (retain the option so that Kokkos::View has the chance to turn it off if it affects performance)
- [X] `std::constant_wrapper`
    - [X] Remove the `constant_wrapper` back-port, as Back-port is only useful for a small range of very recent GCC and Clang implementations
    - [X] Test against GCC trunk with `-std=c++26`; we know it implements `std::constant_wrapper`, because [Compiler Explorer's version of GCC trunk has it](https://godbolt.org/z/3e4qvEvv1).
- [X] Drive-by P2389 (`dims`) fix has already been applied by PR https://github.com/kokkos/mdspan/pull/447 .  That PR will remove import of `dims` into the `experimental` namespace.  This PR should be changed not to do that.  (There's no promise of backwards compatibility for the `experimental` namespace.)
- [ ] Why is `MDSPAN_ENABLE_P3663` not defined in the code?

    - For whatever reason, CMake logic wasn't defining it in the tests.  Deal with this by defining the macro in config.h.  If we don't keep the no-P3663 branch, we can just delete the macro entirely.

- [ ] Decide whether to keep the no-P3663 branch
 
Feature is ON by default.  To disable it, set the CMake option `MDSPAN_ENABLE_P3663` to `OFF`.

Over various commits we've simplified the implementation so that it works with C++17 and subsequent C++ versions.  P3663 relies on `std::constant_wrapper`; we've tested with a version of GCC that implements it, so we can say with some confidence that it "forward-ports" all the way to C++26.


mhoemmen:
@youyu3 Here is the branch in question.

include/experimental/__p2630_bits/strided_slice.hpp
```diff
@@ -40,9 +59,14 @@ struct strided_slice {
...
 
-  static_assert(std::is_integral_v<OffsetType> || detail::mdspan_is_integral_constant<OffsetType>::value);
-  static_assert(std::is_integral_v<ExtentType> || detail::mdspan_is_integral_constant<ExtentType>::value);
-  static_assert(std::is_integral_v<StrideType> || detail::mdspan_is_integral_constant<StrideType>::value);
+  static_assert(detail::__mdspan_is_index_like_v<OffsetType>);
+  static_assert(detail::__mdspan_is_index_like_v<ExtentType>);
+  static_assert(detail::__mdspan_is_index_like_v<StrideType>);
 };
 
+#if (__cplusplus < 202002L)
```
mhoemmen: Consider removing this deduction guide.

include/experimental/__p2630_bits/strided_slice.hpp
```diff
@@ -17,17 +17,36 @@
...
+
 #include <type_traits>
 
 namespace MDSPAN_IMPL_STANDARD_NAMESPACE {
 
 namespace detail {
+
+#if defined(MDSPAN_ENABLE_P3663)
+  template<class T, T Value>
+  using mdspan_constant_wrapper = decltype(cw<Value>);
```
mhoemmen: Consider removing `mdspan_constant_wrapper`.

include/experimental/__p2630_bits/strided_slice.hpp
```diff
@@ -40,9 +53,14 @@ struct strided_slice {
...
 
-  static_assert(std::is_integral_v<OffsetType> || detail::mdspan_is_integral_constant<OffsetType>::value);
-  static_assert(std::is_integral_v<ExtentType> || detail::mdspan_is_integral_constant<ExtentType>::value);
-  static_assert(std::is_integral_v<StrideType> || detail::mdspan_is_integral_constant<StrideType>::value);
+  static_assert(detail::__mdspan_is_index_like_v<OffsetType>);
+  static_assert(detail::__mdspan_is_index_like_v<ExtentType>);
+  static_assert(detail::__mdspan_is_index_like_v<StrideType>);
 };
 
+#if (__cplusplus < 202002L)
```
mhoemmen: Consider using mdspan's macros for C++ version instead of `__cplusplus`.

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -17,6 +17,7 @@
 #pragma once
 
 #include <complex>
+#include <tuple>
```
mhoemmen, line 20: Why do we need this?

tests/test_dims.cpp
```diff
@@ -31,7 +31,7 @@ static constexpr bool is_extents_v<
 template<std::size_t Rank>
 void test_dims_with_one_template_argument()
 {
-  using d = MDSPAN_IMPL_STANDARD_NAMESPACE :: MDSPAN_IMPL_PROPOSED_NAMESPACE :: dims<Rank>;
+  using d = MDSPAN_IMPL_STANDARD_NAMESPACE :: dims<Rank>;
```
mhoemmen, line 34: This is part of the P2389 drive-by fix.

include/experimental/__p2630_bits/submdspan_extents.hpp
mhoemmen: The header should have included `<tuple>` before, because there were specializations of `index_pair_like` for `std::tuple`.

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -85,28 +128,106 @@ struct index_pair_like<std::array<IdxT, 2>, IndexType> {
...
+    ! std::is_unsigned_v<Integral> &&
+    (
+      std::is_convertible_v<Integral, size_t> ||
+      std::is_convertible_v<Integral, int>
+    )
+  )
+)
+MDSPAN_INLINE_FUNCTION
+constexpr Integral first_of(const Integral &i) {
+  // FIXME (mfh 2025/06/06) This is broken, but it's better than it was.
```
mhoemmen, line 190: We should just fix the non-P3663 branch if we intend to keep it.

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -85,28 +128,106 @@ struct index_pair_like<std::array<IdxT, 2>, IndexType> {
...
+
+#if defined(MDSPAN_ENABLE_P3663)
+template<auto Value>
+MDSPAN_INLINE_FUNCTION
+constexpr constant_wrapper<Value>
+first_of(constant_wrapper<Value>) {
+  return {};
+}
+#else
+// NOTE This is technically not conforming.
```
mhoemmen, line 206: We should just fix this.

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -150,49 +273,105 @@ first_of(const strided_slice<OffsetType, ExtentType, StrideType> &r) {
...
+  const Extents&,
+  const Integral& i)
+{
   return i;
 }
 
+// clang++ with C++14 is not fond of the pragma appearing inside the
+// macro definition.  In that case, it complains, "error: embedding a
+// directive within macro arguments has undefined behavior
+// [-Werror,-Wembedded-directive]."  The fix is to duplicate code.
```
mhoemmen: This describes code that was later removed.  I think we can delete this comment.

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -268,6 +479,21 @@ constexpr auto divide(const T0 &v0, const T1 &v1) {
   return IndexT(v0) / IndexT(v1);
 }
 
+#if defined(MDSPAN_ENABLE_P3663)
+template <class IndexType, auto v0, auto v1>
+MDSPAN_INLINE_FUNCTION
+constexpr auto divide(constant_wrapper<v0> i0,
```
mhoemmen, line 470: Very few compilers actually support `constant_wrapper` arithmetic yet.  That's why we have these functions.

(EDIT: There's nothing to fix here; I'm just pointing this out to reviewers.)

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -404,6 +681,571 @@ struct extents_constructor<0, Extents, NewStaticExtents...> {
...
+
+// DONE Added to P3663R2: Check wording of check-static-bounds
+// so that it only assumes that types are default constructible
+// in constant expressions if they are integral-constant-like.
+
+template<size_t k, class S_k, class IndexType, size_t... Exts>
+  constexpr check_static_bounds_result check_static_bounds(
+    const extents<IndexType, Exts...>&)
+{
+#if defined(__cpp_pack_indexing)
```
mhoemmen, line 813: It might make sense just to remove the branch and define a function for getting the k-th type of a pack.  (See `get_kth_in_pack` below.)

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -404,6 +681,571 @@ struct extents_constructor<0, Extents, NewStaticExtents...> {
...
+// sense because the function only tests one slice (the k-th one).
+// Also, taking slices parameter(s) makes use of check_static_bounds
+// not a constant expression.
+
+// DONE Added to P3663R2: Check wording of check-static-bounds
+// so that it only assumes that types are default constructible
+// in constant expressions if they are integral-constant-like.
+
+template<size_t k, class S_k, class IndexType, size_t... Exts>
+  constexpr check_static_bounds_result check_static_bounds(
```
mhoemmen, line 810: FIXME In R3, this returns `bool` instead of an enum.

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -404,6 +681,571 @@ struct extents_constructor<0, Extents, NewStaticExtents...> {
...
+MDSPAN_INLINE_FUNCTION
+constexpr void
+check_canonical_kth_subdmspan_slice_types(
+  const extents<IndexType, Extents...>& exts, Slices... slices)
+{
+#if (__cplusplus < 202002L)
+  check_canonical_kth_subdmspan_slice_types_impl(
+    std::make_index_sequence<sizeof...(Slices)>(), exts, slices...);
+#else
+  // We really want to keep the C++20 branch here
```
mhoemmen: Consider removing this branch.  It's harder to maintain two implementations.  Kokkos requires C++20 or greater, so the pre-C++20 branch wouldn't be tested as often if it's kept separate.

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -404,6 +681,571 @@ struct extents_constructor<0, Extents, NewStaticExtents...> {
...
+
+  if constexpr (std::is_convertible_v<Slice, full_extent_t>) {
+    return full_extent; // 11.1
+  }
+  else if constexpr (std::is_convertible_v<Slice, IndexType>) {
+    return canonical_ice<IndexType>(s); // 11.2
+  }
+  else if constexpr (is_strided_slice<Slice>::value) { // 11.3
+#if ! defined(MDSPAN_CONSTANT_WRAPPER_WORKAROUND)
+    // GCC 11.4.0 (C++20) accepts this code, but Clang 14 does not.
```
mhoemmen: Just delete this branch.  icpx also doesn't like this (and its feature test macros don't prevent warnings-as-errors).

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -404,6 +681,571 @@ struct extents_constructor<0, Extents, NewStaticExtents...> {
...
+      /* .extent = */ extent,
+      /* .stride = */ stride
+    };
+#endif
+  }
+#if ! defined(__cpp_lib_tuple_like) || (__cpp_lib_tuple_like < 202311L)
+  else if constexpr (detail::is_std_complex<Slice>) {
+#if ! defined(MDSPAN_CONSTANT_WRAPPER_WORKAROUND)
+    // GCC 11.4.0 (C++20) accepts this code, but Clang 14 does not.
+    return strided_slice{
```
mhoemmen: Just delete this branch.

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -404,6 +681,571 @@ struct extents_constructor<0, Extents, NewStaticExtents...> {
...
+#endif
+  else { // 11.4
+    auto [s_k0, s_k1] = s;
+    using S_k0 = decltype(s_k0);
+    using S_k1 = decltype(s_k1);
+    static_assert(std::is_convertible_v<S_k0, IndexType>);
+    static_assert(std::is_convertible_v<S_k1, IndexType>);
+#if ! defined(MDSPAN_CONSTANT_WRAPPER_WORKAROUND)
+    // GCC 11.4.0 (C++20) accepts this code, but Clang 14 does not.
+    return strided_slice{
```
mhoemmen: Just delete this branch.

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -404,6 +681,571 @@ struct extents_constructor<0, Extents, NewStaticExtents...> {
...
+  )
+)
+MDSPAN_INLINE_FUNCTION
+constexpr auto
+submdspan_canonicalize_slices_impl(
+  std::index_sequence<Inds...>,
+  const extents<IndexType, Extents...>& exts,
+  Slices... slices)
+{
+  return std::tuple{
```
mhoemmen, line 1158: Consider using `Kokkos::tuple` instead of `std::tuple`.

include/experimental/__p2630_bits/submdspan_mapping.hpp
```diff
@@ -60,10 +139,42 @@ namespace detail {
...
 one_slice_out_of_bounds(const IndexType &ext, const Slice &slice) {
+#if defined(MDSPAN_ENABLE_P3663)
   using common_t =
-      std::common_type_t<decltype(detail::first_of(slice)), IndexType>;
-  return static_cast<common_t>(detail::first_of(slice)) ==
+      std::common_type_t<decltype(first_of(slice)), IndexType>;
+  return static_cast<common_t>(first_of(slice)) ==
          static_cast<common_t>(ext);
+#else
+  // NOTE (mfh 2025/06/06) The original implementation was not conforming.
```
mhoemmen, line 148: We should decide whether to fix the non-P3663 case or just remove it and make P3663 the only implementation.

tests/test_strided_slice.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include <type_traits>
+
+#include <gtest/gtest.h>
+
+namespace {
+
+template<class OffsetType, class ExtentType, class StrideType>
+void test_strided_slice(OffsetType offset, ExtentType extent, StrideType stride)
+{
+  // Some compilers are bad at CTAD for aggregates.
```
mhoemmen, line 26: We should consider branching based on compilers where it's known to work.  It would be good to test that CTAD works.

benchmarks/submdspan/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+
+mdspan_add_benchmark(submdspan)
+
+#if(MDSPAN_ENABLE_CUDA)
+#  add_subdirectory(cuda)
+#endif()
```
nmm0:
```suggestion
```

include/experimental/__p2630_bits/constant_wrapper.hpp
```diff
@@ -0,0 +1,265 @@
+#pragma once
```
nmm0, line 1: I think we should pull this into a separate PR.

Also since we are pulling this in should we ask Zach if it's under a license or public domain? I couldn't find any info in the source code so I suspect it is the latter.

mhoemmen:
PR https://github.com/kokkos/mdspan/pull/447 will remove import of `dims` into the `experimental` namespace.  This PR should be changed not to do that.  (There's no promise of backwards compatibility for the `experimental` namespace.)

include/experimental/__p2630_bits/remove_cvref.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include <type_traits>
+
+namespace MDSPAN_IMPL_STANDARD_NAMESPACE {
```
nmm0, line 5: I think this is fine here but at some point if we reorganize files I think we could move all of these little backports of C++ things that we need into its own utility folder.

include/experimental/__p2630_bits/strided_slice.hpp
```diff
@@ -17,17 +17,30 @@
...
+  
+  template<auto Value, class Type>
+  constexpr bool is_constant_wrapper<constant_wrapper<Value, Type>> = true;  
+#endif // MDSPAN_ENABLE_P3663
 
-  template<class T, T val>
-  struct mdspan_is_integral_constant<std::integral_constant<T,val>>: std::true_type {};
-}
+  template<class T>
+  constexpr bool __mdspan_is_index_like_v =
```
nmm0, line 46: Just trying to wrap my brain around this -- is this the same behavior as before?

include/experimental/__p2630_bits/strided_slice.hpp
```diff
@@ -40,9 +53,14 @@ struct strided_slice {
...
-  static_assert(std::is_integral_v<ExtentType> || detail::mdspan_is_integral_constant<ExtentType>::value);
-  static_assert(std::is_integral_v<StrideType> || detail::mdspan_is_integral_constant<StrideType>::value);
+  static_assert(detail::__mdspan_is_index_like_v<OffsetType>);
+  static_assert(detail::__mdspan_is_index_like_v<ExtentType>);
+  static_assert(detail::__mdspan_is_index_like_v<StrideType>);
 };
 
+#if (__cplusplus < 202002L)
+template <class OffsetType, class ExtentType, class StrideType>
+strided_slice(const OffsetType&, const ExtentType&, const StrideType&) ->
```
nmm0: Is this deduction guide needed? I know you said to consider removing it but it seems like this shouldn't do anything unless I'm missing something

include/experimental/__p2630_bits/submdspan.hpp
```diff
@@ -20,12 +20,58 @@
...
 template <class ElementType, class Extents, class LayoutPolicy,
           class AccessorPolicy, class... SliceSpecifiers>
 MDSPAN_INLINE_FUNCTION
 constexpr auto
 submdspan(const mdspan<ElementType, Extents, LayoutPolicy, AccessorPolicy> &src,
           SliceSpecifiers... slices) {
+
+#if defined(MDSPAN_ENABLE_P3663)
+
+#  if defined(__cpp_structured_bindings) && (__cpp_structured_bindings >= 202411L)
```
nmm0: To be honest unless this really improves compile times I'd just stick with the tuple implementation so we only have one implementation to maintain/understand

include/experimental/__p2630_bits/submdspan.hpp
```diff
@@ -20,12 +20,58 @@
...
+  auto sub_map_result =
+    submdspan_mapping(src.mapping(), canonical_slices...);
+
+#  else
+
+  auto canonical_slices_tuple =
+    submdspan_canonicalize_slices(src.extents(), slices...);
+  using src_mapping_type = decltype(src.mapping()); // CTAD doesn't seem to work
+  auto sub_map_result = std::apply(
+    detail::submdspan_mapping_caller<src_mapping_type>{src.mapping()},
```
nmm0: This might be a little more readable as a lambda and we wouldn't need the template argument since mapping would be captured. This way we don't have to jump around to see the definition

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -85,28 +128,106 @@ struct index_pair_like<std::array<IdxT, 2>, IndexType> {
...
+#endif // ! defined(MDSPAN_ENABLE_P3663)
+
 // first_of(slice): getting begin of slice specifier range
+
+#if defined(MDSPAN_ENABLE_P3663)
+
 MDSPAN_TEMPLATE_REQUIRES(
   class Integral,
-  /* requires */(std::is_convertible_v<Integral, size_t>)
+  /* requires */(std::is_signed_v<Integral> || std::is_unsigned_v<Integral>)
```
nmm0: Does this actually need any constraints specifying it is integral? Or it would never get to the point where its called by a float or something

CMakeLists.txt
```diff
@@ -28,10 +28,15 @@ option(MDSPAN_GENERATE_STD_NAMESPACE_TARGETS "Whether to generate and install ta
 
 # Option to override which C++ standard to use
 set(MDSPAN_CXX_STANDARD DETECT CACHE STRING "Override the default CXX_STANDARD to compile with.")
-set_property(CACHE MDSPAN_CXX_STANDARD PROPERTY STRINGS DETECT 14 17 20 23)
+set_property(CACHE MDSPAN_CXX_STANDARD PROPERTY STRINGS DETECT 14 17 20 23 26)
 
 option(MDSPAN_ENABLE_CONCEPTS "Try to enable concepts support by giving extra flags." On)
 
+option(MDSPAN_ENABLE_P3663 "Enable implementation of P3663 (Future-proof submdspan_mapping)." On)
```
nmm0, line 35: I think this is actually our only cmake-based feature switch? So this actually leads to the question of how we do this, because right now if someone is using mdspan via cmake this does nothing (as it only enables a compile definition for benchmarks).

So I suppose the question is do we want to add them to a target (e.g. via target_compile_definitions) or via a `#cmakedefine` sort of thing like Kokkos does where we generate a header with the define. We could even let it be overrideable via a define.

I actually am leaning towards the latter because we could use it in our single-header mode and the value could be overriden if using the header only mode via a define.

nmm0 (commented):
Ok some initial comments, I got through a good chunk of it but not everything.

include/experimental/__p2630_bits/constant_wrapper.hpp
mhoemmen: @nmm0 I'm inclined to check whether any (VERY VERY NEW) compiler comes with a `constant_wrapper` implementation.  If it does, we can test the code with the actual `std::constant_wrapper`, remove the back-port, and leave only the pre-C++26 work-around (that looks like `integral_constant`).  What do you think?

include/experimental/__p2630_bits/remove_cvref.hpp
mhoemmen: That's a good idea!  If we end up removing the `constant_wrapper` back-port (please see my comment above), then there won't be quite so many long files here.

include/experimental/__p2630_bits/strided_slice.hpp
mhoemmen: @nmm0 It's not the same behavior as before.  The existing implementation is NOT conforming.  Even before P3663, _`integral-constant-like`_ is a concept that user-defined types can meet.  However, the code only called `std::integral_constant` and `Kokkos::integral_constant` "integral constant like."

include/experimental/__p2630_bits/strided_slice.hpp
mhoemmen: It didn't actually help with icpx (which is a Clang derivative if I remember right).  We should remove it; aggregates like this shouldn't need a deduction guide.  For broken compilers, users would have to work around for other things anyway.

CMakeLists.txt
mhoemmen: @nmm0 wrote:
> ... because right now if someone is using mdspan via cmake this does nothing....

Practically, it changes the implementation by changing the macro definitions in the headers -- at least it should!  I might be missing something here.

Should we consider just removing this option and making the library implement P3663 unconditionally?

include/experimental/__p2630_bits/strided_slice.hpp
mhoemmen: I've removed the deduction guide for `strided_slice`.

mhoemmen:
Compiler Explorer's version of GCC trunk has `std::constant_wrapper`: https://godbolt.org/z/3e4qvEvv1 .

include/experimental/__p2630_bits/strided_slice.hpp
mhoemmen: I've removed `strided_slice`'s deduction guide.

tests/test_dims.cpp
mhoemmen: I've fixed merge conflicts with `stable` as a result of merging PR #447 .

include/experimental/__p2630_bits/constant_wrapper.hpp
mhoemmen: Compiler Explorer's GCC trunk has `std::constant_wrapper`: https://godbolt.org/z/3e4qvEvv1 .  That means we have a way to test this with the actual `std::constant_wrapper`.  That motivates removing the back-port, as long as we know that it actually works with `std::constant_wrapper`.

C++26 may still see changes to `std::constant_wrapper` -- there's a fix paper in flight -- but I don't think it would apply to anything in this PR.  (The paper would change behavior for callables.)

include/experimental/__p2630_bits/submdspan.hpp
mhoemmen: I've removed the C++26 branch.

include/experimental/__p2630_bits/submdspan.hpp
mhoemmen: I've reimplemented this as a lambda with `auto&&... canonical_slices` capture.  Thanks!

include/experimental/__p2630_bits/submdspan_extents.hpp
mhoemmen: P3663 means that `first_of` only ever sees

1. `full_extent_t`,
2. `index_type` or _`integral-constant-like`_ of `index_type`, or
3. `strided_slice` with template parameters satisfying (2).

I've simplified this a bit to make that more clear, and to remove the need for `requires` clauses on `first_of` in the P3663 case.  (I just use function overloads.)

include/experimental/__p2630_bits/strided_slice.hpp
mhoemmen: I've fixed `__mdspan_is_index_like_v` so that it's now conforming even in the no-P3663 case.  There are other places in which the no-P3663 case is still nonconforming, though.  The comments in the code should point those out.

include/experimental/__p2630_bits/strided_slice.hpp
mhoemmen: I removed it.

include/experimental/__p2630_bits/strided_slice.hpp
mhoemmen: I don't use this macro any more.

include/experimental/__p2630_bits/submdspan_extents.hpp
mhoemmen: I've deleted this comment.

include/experimental/__p2630_bits/constant_wrapper.hpp
nmm0: It does look like GCC16 will have `std::constant_wrapper` (so i guess it made the cut): https://gcc.gnu.org/gcc-16/changes.html

I'm fine with having a constant wrapper backport though, just will be easier to review in another PR

include/experimental/__p2630_bits/strided_slice.hpp
nmm0: ohhh I see. Thanks for fixing it

mhoemmen:
Hi @nmm0 !  GCC trunk ("16.0.1") has `std::constant_wrapper`.  Commit 971c8fb6068bcebf4399cf4a18e8133633c0ec91 builds and passes tests with that, using the following cmake options.

```
cmake ${SOURCE_DIRECTORY} \
  -DMDSPAN_ENABLE_TESTS=ON \
  -DMDSPAN_ENABLE_EXAMPLES=ON \
  -DMDSPAN_ENABLE_P3663=ON \
  -DCMAKE_CXX_FLAGS="-Wall -std=c++26" \
  -DCMAKE_CXX_COMPILER=${GCC_TRUNK_PREFIX}/bin/g++ \
  -DMDSPAN_CXX_STANDARD=26
```

As a result, I think it's safe to remove the `constant_wrapper` back-port entirely.

include/experimental/__p2630_bits/submdspan_extents.hpp
mhoemmen: I've done this.

include/experimental/__p2630_bits/submdspan_extents.hpp
mhoemmen: I've removed this branch.

include/experimental/__p2630_bits/submdspan_extents.hpp
mhoemmen: I've removed this branch.

include/experimental/__p2630_bits/submdspan_extents.hpp
mhoemmen: I've removed this branch.

include/experimental/__p2630_bits/constant_wrapper.hpp
mhoemmen: @nmm0 The back-port is really only useful for Clang 21 and just barely for GCC 15.  Anything earlier won't have a chance of building a `constant_wrapper` that looks enough like `constant_wrapper` to call it a back-port.

I've removed the back-port.  GCC trunk builds it correctly.  I'm not convinced that `MDSPAN_ENABLE_P3663` is always defined in the build, though.

benchmarks/submdspan/cuda/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+
+if(CMAKE_CUDA_COMPILER_ID STREQUAL "NVIDIA")
+  set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} --extended-lambda")
```
mhoemmen, line 3: We need to test with this flag removed.

benchmarks/submdspan/CMakeLists.txt
mhoemmen: Fixed; thanks! : - )

include/experimental/__p0009_bits/config.hpp
```diff
@@ -303,3 +303,7 @@ static_assert(MDSPAN_IMPL_CPLUSPLUS >= MDSPAN_CXX_STD_14, "mdspan requires C++14
 #  define MDSPAN_IMPL_OP5(mds, a, b, c, d, e) mds(a,b,c,d,e)
 #  define MDSPAN_IMPL_OP6(mds, a, b, c, d, e, f) mds(a,b,c,d,e,f)
 #endif
+
+#if ! defined(MDSPAN_ENABLE_P3663)
+#  define MDSPAN_ENABLE_P3663 1
```
mhoemmen, line 308: We may want to revisit this if we decide we want to keep the no-P3663 branch.

include/experimental/__p2630_bits/constant_wrapper.hpp
mhoemmen: I've unconditionally defined the macro `MDSPAN_ENABLE_P3663` in config.h (see above) and GCC trunk builds everything.

include/experimental/__p2630_bits/remove_cvref.hpp
mhoemmen: I ended up removing the `constant_wrapper` back-port, btw.

include/experimental/__p2630_bits/submdspan_extents.hpp
mhoemmen: This requires replacing use of `std::apply` in `submdspan`.

include/experimental/__p2630_bits/submdspan_extents.hpp
mhoemmen: I did : - )

PR #410, opened 2025-06-02
precondition checks in extents constructors

nmm0:
This adds the missing precondition checks for [mdspan.extents.cons](https://eel.is/c++draft/views.multidim#mdspan.extents.cons)

### DIFF
--- a/include/experimental/__p0009_bits/extents.hpp
+++ b/include/experimental/__p0009_bits/extents.hpp
@@ -439,7 +439,10 @@ public:
            sizeof...(OtherIndexTypes) == m_rank_dynamic)))
   MDSPAN_INLINE_FUNCTION
   constexpr explicit extents(OtherIndexTypes... dynvals) noexcept
-      : m_vals(static_cast<index_type>(dynvals)...) {}
+      : m_vals(static_cast<index_type>(dynvals)...) {
+    MDSPAN_IMPL_PRECONDITION(
+        detail::all_values_are_nonnegative_and_representable<index_type>(dynvals...));
+  }
 
   MDSPAN_TEMPLATE_REQUIRES(
       class OtherIndexType, size_t N,
@@ -452,7 +455,11 @@ public:
   MDSPAN_INLINE_FUNCTION
   MDSPAN_CONDITIONAL_EXPLICIT(N != m_rank_dynamic)
   constexpr extents(const std::array<OtherIndexType, N> &exts) noexcept
-      : m_vals(std::move(exts)) {}
+      : m_vals(std::move(exts)) {
+    MDSPAN_IMPL_PRECONDITION(
+        detail::range_is_nonnegative_and_representable<index_type>(
+            std::begin(exts), std::end(exts)));
+  }
 
 #ifdef __cpp_lib_span
   MDSPAN_TEMPLATE_REQUIRES(
@@ -464,7 +471,11 @@ public:
   MDSPAN_INLINE_FUNCTION
   MDSPAN_CONDITIONAL_EXPLICIT(N != m_rank_dynamic)
   constexpr extents(const std::span<OtherIndexType, N> &exts) noexcept
-      : m_vals(std::move(exts)) {}
+      : m_vals(std::move(exts)) {
+    MDSPAN_IMPL_PRECONDITION(
+        detail::range_is_nonnegative_and_representable<index_type>(
+            std::begin(exts), std::end(exts)));
+  }
 #endif
 
 private:
@@ -536,10 +547,14 @@ public:
                                ...) ||
                               (std::numeric_limits<index_type>::max() <
                                std::numeric_limits<OtherIndexType>::max()))
-  constexpr extents(const extents<OtherIndexType, OtherExtents...> &other) noexcept
+  constexpr extents(
+      const extents<OtherIndexType, OtherExtents...> &other) noexcept
       : m_vals(impl_construct_vals_from_extents(
             std::integral_constant<size_t, 0>(),
-            std::integral_constant<size_t, 0>(), other)) {}
+            std::integral_constant<size_t, 0>(), other)) {
+    MDSPAN_IMPL_PRECONDITION(
+        detail::extent_is_representable<index_type>(other));
+  }
 
   // Comparison operator
   template <class OtherIndexType, size_t... OtherExtents>
--- a/include/experimental/__p0009_bits/extents.hpp
+++ b/include/experimental/__p0009_bits/extents.hpp
@@ -440,8 +440,10 @@ public:
   MDSPAN_INLINE_FUNCTION
   constexpr explicit extents(OtherIndexTypes... dynvals) noexcept
       : m_vals(static_cast<index_type>(dynvals)...) {
+#if MDSPAN_HAS_CXX_17
     MDSPAN_IMPL_PRECONDITION(
         detail::all_values_are_nonnegative_and_representable<index_type>(dynvals...));
+#endif
   }
 
   MDSPAN_TEMPLATE_REQUIRES(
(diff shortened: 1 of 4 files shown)


tests/test_extents.cpp
```diff
@@ -481,3 +481,53 @@ TEST(TestExtentsCTADStdArray, test_extents_ctad_std_array) {
 }
 */
 #endif
+
+#if MDSPAN_HAS_CXX_17
```
dalg24, line 485: Also if preconditions are checked

tests/test_extents.cpp
```diff
@@ -481,3 +481,53 @@ TEST(TestExtentsCTADStdArray, test_extents_ctad_std_array) {
...
+TEST(TestExtentsConstructorPreconditions, test_extents_construct_indices) {
+  auto test_precondition_indices_not_representable = [] {
+    [[maybe_unused]] auto exts = Kokkos::dextents< std::int8_t, 2 >{ 500, 500 };
+  };
+  EXPECT_DEATH(test_precondition_indices_not_representable(), "");
```
dalg24, line 491: We do not want to ensure that the error message matches some regex?

tests/test_extents.cpp
dalg24: OK I saw the generator expression for the compile definition.

include/experimental/__p0009_bits/utility.hpp
```diff
@@ -198,10 +198,70 @@ MDSPAN_INLINE_FUNCTION constexpr bool cmp_greater_equal(T t, U u) noexcept {
 
 template <class R, class T>
 MDSPAN_INLINE_FUNCTION constexpr bool in_range(T t) noexcept {
+  static_assert(std::is_integral_v<R> && std::is_integral_v<T>);
```
dalg24, line 206: Sure

include/experimental/__p0009_bits/utility.hpp
```diff
@@ -198,10 +198,70 @@ MDSPAN_INLINE_FUNCTION constexpr bool cmp_greater_equal(T t, U u) noexcept {
...
+  {
+    if constexpr (std::is_signed_v<T>) {
+      if (t < 0)
+        return false;
+    }
+
+    return in_range<R>(t);
+  } else
+  {
+    if constexpr (std::is_signed_v<T>) {
```
dalg24: Did you mean to say if `R` is signed?

include/experimental/__p0009_bits/utility.hpp
```diff
@@ -198,10 +198,70 @@ MDSPAN_INLINE_FUNCTION constexpr bool cmp_greater_equal(T t, U u) noexcept {
...
+    }
+
+    return true;
+  }
+}
+
+template<class R, class... Values>
+MDSPAN_INLINE_FUNCTION constexpr bool
+all_values_are_representable(Values... values) noexcept {
+  return ( in_range<R>( values ) && ... && true );
```
dalg24: The `&& true` is unnecessary.  I would drop it.  Same comment below.

nmm0: thought I had removed them, will remove

tests/test_extents.cpp
nmm0: We can at least for CPU but it won't work with CUDA. So I just decided not to since it doesn't matter a whole lot

tests/test_extents.cpp
nmm0: I can still add the precondition enabled macro if you want?

tests/test_extents.cpp
dalg24: What do you mean it won't work with CUDA?
We do check error messages from the device in Kokkos so there is some level of support.

include/experimental/__p0009_bits/extents.hpp
```diff
@@ -439,7 +439,12 @@ template <class IndexType, size_t... Extents> class extents {
            sizeof...(OtherIndexTypes) == m_rank_dynamic)))
   MDSPAN_INLINE_FUNCTION
   constexpr explicit extents(OtherIndexTypes... dynvals) noexcept
-      : m_vals(static_cast<index_type>(dynvals)...) {}
+      : m_vals(static_cast<index_type>(dynvals)...) {
+#if MDSPAN_HAS_CXX_17
+    MDSPAN_IMPL_PRECONDITION(
+        detail::all_values_are_nonnegative_and_representable<index_type>(dynvals...));
+#endif
```
dalg24, line 446: Just noting that the precondition check will be performed *after* the data member is initialized.

include/experimental/__p0009_bits/utility.hpp
```diff
@@ -198,10 +198,70 @@ MDSPAN_INLINE_FUNCTION constexpr bool cmp_greater_equal(T t, U u) noexcept {
...
+template<class R, class... Values>
+MDSPAN_INLINE_FUNCTION constexpr bool
+all_values_are_nonnegative_and_representable(Values... values) noexcept {
+  return ( is_nonnegative_and_representable<R>( values ) && ... );
+}
+
+template<class R, class ContiguousIterator>
+MDSPAN_INLINE_FUNCTION constexpr bool
+range_is_nonnegative_and_representable(ContiguousIterator begin, ContiguousIterator end) noexcept {
+  for ( auto it = begin; it != end; ++it )
```
dalg24: Contiguous iterators is a refinement of random access
I would assert that begin is less equal than end or at the very least compare `it < end`

include/experimental/__p0009_bits/extents.hpp
nmm0: Yeah, I think it's too complex to check this before initialization -- this is also the approach libc++ takes so I think it's okay

tests/test_extents.cpp
nmm0: So for now the CUDA precondition violation handler prints to stdout. `EXPECT_DEATH` checks stderr -- we could do something like:

```cpp
#if defined(MDSPAN_IMPL_HAS_CUDA) || defined(MDSPAN_IMPL_HAS_HIP) || defined(MDSPAN_IMPL_HAS_SYCL)
  EXPECT_DEATH(test_precondition_indices_not_representable(), "");
#else
  EXPECT_DEATH(test_precondition_indices_not_representable(), "extent_is_representable");
#endif
```

We do this elsewhere. But really the precondition handler doesn't print the best errors anyway

nmm0: I updated to test the message on host. On device the message is kinda useless:

```
void Kokkos::detail::default_precondition_violation_handler(const char*, const char*, unsigned int): Assertion `0' failed.
```

So I think we should avoid testing device messages for now until we update our precondition violation handler on device (likely a separate PR)

dalg24: Oh, I see, because we call `printf(...)` and `assert(false)` instead of `__assert_fail(...)` like we do in Kokkos

tests/test_extents.cpp
```diff
@@ -481,3 +481,81 @@ TEST(TestExtentsCTADStdArray, test_extents_ctad_std_array) {
 }
 */
 #endif
+
+#if MDSPAN_HAS_CXX_17
+TEST(TestExtentsConstructorPreconditions, test_extents_construct_indices) {
+  auto test_precondition_indices_not_representable = [] {
+    [[maybe_unused]] auto exts = Kokkos::dextents< std::int8_t, 2 >{ 500, 500 };
```
dalg24, line 488: Why do we have these maybe unused attributes here and below?

nmm0: exts isn't used here just the constructor preconditions are called (and will fault, hence the death tests below). So the compiler will warn

dalg24 (approved):

crtrott (approved):

PR #285, opened 2023-10-20
Fix standard compliance

crtrott:
This fixes most everything identified in the llvm tests which is wrong with our reference implementation, and where we haven't backported fixes yet from when we contributed our implementation to LLVM. 

I haven't done mdspan yet.

PR #413, opened 2025-06-12
Avoid TU-local entities

masterleinad:
Related to https://github.com/kokkos/kokkos/pull/8117. While trying to build a C++20 module for Kokkos with g++-15, I ran into a couple of instances where TU-local entities (https://en.cppreference.com/w/cpp/language/tu_local.html, https://eel.is/c++draft/basic.link#14) are exposed. This pull request tries to fix those I was able to address.
Note that clang++ is much more lenient with TU-local entities as decribed at https://clang.llvm.org/docs/StandardCPlusPlusModules.html#using-tu-local-entity-in-other-units.

- anonymous structs are TU-local even if the respective variable is declared `inline`.
- anonymous namespaces create TU-local symbols
- It's not quite clear to me why `compute_s_static_layout_left` and `compute_s_static_layout_right` were considered TU-local but converting those structs to functions seems be an improvement (since it's more concise) anyway.

~~It's not quite clear to me which clang-format version was used for these files before so I kept the indentation fixes with clang-16 in a separate commit.~~

### DIFF
--- a/include/experimental/__p0009_bits/utility.hpp
+++ b/include/experimental/__p0009_bits/utility.hpp
@@ -50,7 +50,7 @@ constexpr bool rankwise_equal(with_rank<N>, const T1& x, const T2& y, F func)
 #if MDSPAN_HAS_CXX_17
 inline
 #endif
-constexpr struct
+constexpr struct extent_struct
 {
   template <class T, class I>
   MDSPAN_INLINE_FUNCTION
@@ -63,7 +63,7 @@ constexpr struct
 #if MDSPAN_HAS_CXX_17
 inline
 #endif
-constexpr struct
+constexpr struct stride_struct
 {
   template <class T, class I>
   MDSPAN_INLINE_FUNCTION
--- a/include/experimental/__p0009_bits/utility.hpp
+++ b/include/experimental/__p0009_bits/utility.hpp
@@ -50,7 +50,7 @@ constexpr bool rankwise_equal(with_rank<N>, const T1& x, const T2& y, F func)
 #if MDSPAN_HAS_CXX_17
 inline
 #endif
-constexpr struct extent_struct
+constexpr struct extent_functor
 {
   template <class T, class I>
   MDSPAN_INLINE_FUNCTION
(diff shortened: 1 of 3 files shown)


masterleinad:
https://github.com/kokkos/mdspan/pull/285#discussion_r1369043961 suggests that the code was never formatted with clang-format although there are some clang-format comments in the code. Let me just drop the last commit.

nmm0:
> [#285 (comment)](https://github.com/kokkos/mdspan/pull/285#discussion_r1369043961) suggests that the code was never formatted with clang-format although there are some clang-format comments in the code. Let me just drop the last commit.

yeah at some point we need to run the repo through clang-format and add a check to enforce it

include/experimental/__p0009_bits/utility.hpp
```diff
@@ -50,7 +50,7 @@ constexpr bool rankwise_equal(with_rank<N>, const T1& x, const T2& y, F func)
 #if MDSPAN_HAS_CXX_17
 inline
 #endif
-constexpr struct
+constexpr struct extent_struct
```
nmm0:
```suggestion
constexpr struct extent_functor
```

include/experimental/__p0009_bits/utility.hpp
```diff
@@ -63,7 +63,7 @@ constexpr struct
 #if MDSPAN_HAS_CXX_17
 inline
 #endif
-constexpr struct
+constexpr struct stride_struct
```
nmm0:
```suggestion
constexpr struct stride_functor
```

include/experimental/__p2630_bits/strided_slice.hpp
```diff
@@ -21,13 +21,11 @@
 
 namespace MDSPAN_IMPL_STANDARD_NAMESPACE {
 
-namespace {
```
nmm0, line 24: I think keep the namespace and rename to `detail` so it is not anonymous

masterleinad:
> yeah at some point we need to run the repo through clang-format and add a check to enforce it

I'm happy to open a pull request if you're fine with doing it now.

nmm0 (commented):
A couple of minor suggestions

nmm0:
> > yeah at some point we need to run the repo through clang-format and add a check to enforce it
> 
> I'm happy to open a pull request if you're fine with doing it now.

That would be really helpful if you have the time!

nmm0 (approved):

PR #414, opened 2025-06-12
Add GitHub clang-format-16 check and indent accordingly

masterleinad:
In response to https://github.com/kokkos/mdspan/pull/413#issuecomment-2967400086 and https://github.com/kokkos/mdspan/pull/285#discussion_r1369043961, this pull request enforces the repository to be formatted with clang-format 16 using the same settings as for https://github.com/kokkos/kokkos.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,6 @@
+#Official Tool: clang-format version 16.0.0
+BasedOnStyle: google
+SortIncludes: false
+AlignConsecutiveAssignments: true
+AllowShortCaseLabelsOnASingleLine: true
+AllowShortIfStatementsOnASingleLine: true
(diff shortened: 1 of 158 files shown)


tpadioleau (approved):
yes!

.clang-format
```diff
@@ -0,0 +1,6 @@
+#Official Tool: clang-format version 16.0.0
+BasedOnStyle: google
+SortIncludes: false
+AlignConsecutiveAssignments: true
+AllowShortCaseLabelsOnASingleLine: true
+AllowShortIfStatementsOnASingleLine: true
```
dalg24, line 6: Note that there is no mandate to adopt the same style as Kokkos Core

dalg24: I would recommend picking something that minimizes the diff.

nmm0 (approved):

PR #416, opened 2025-06-16
fix google benchmark deprecated warnings

nmm0:
Google benchmark deprecated the use of `DoNotOptimize` with const ref. According to the deprecation warning, compilers might still optimize const refs. This PR creates temporary variables in order to use the const ref versions.


tpadioleau (approved):
Looks good to me, this is also what was recommended in https://github.com/google/benchmark/issues/1580#issuecomment-1484916108

dalg24 (approved):
Deprecation implemented in https://github.com/google/benchmark/pull/1493

PR #421, opened 2025-06-18
remove `MDSPAN_CXX_STANDARD`

nmm0:
This also reworks the way we set the C++ Standard. I marked this as a draft since I'm not 100% yet we should go this way. There is an alternate approach.

This approach works by setting a required standard per target via `target_compile_features`. The way this works is that if `CMAKE_CXX_STANDARD` is newer, cmake will provide the target the newer C++ standard flags. The idea is that `CMAKE_CXX_STANDARD` should only be provided by the user.

If `CMAKE_CXX_STANDARD` is lower than the required standard of some of the targets, cmake will still use the higher standard flags (assuming the compiler supports them).

The lowest feature flag is `cxx_std_14`; it means that if the user doesn't have a compiler supporting C++14 cmake should error. Some of the tests are predicated on the compiler supporting newer standards.

Finally, I still need to figure out how this should work with CUDA/HIP; thus, this should stay a draft until I can sort that.

### Alternative approach

We can instead check user-provided `CMAKE_CXX_STANDARD` to make sure it's at least 14; if it is not set we can initialize it to the default. This is less verbose, so we set the standard for all of the targets at once. We can still set per-target properties for `CXX_STANDARD`, though if we do that I'd prefer the `target_compile_features` approach.


nmm0:
So a funny thing is that now even the C++14 tests run the C++17 requiring tests (in C++17 mode, though the other non-c++17 requiring tests tests are in C++14) because the compiler supports C++17.

mhoemmen:
It would be excellent if we could update https://github.com/kokkos/stdBLAS once this merges; thanks!

PR #423, opened 2025-08-11
workaround gcc substition bug

nmm0:
Affects versions 11 and 12 (and probably less than those). This bug showed up in Kokkos, where a substitution failure in `is_mapping_of` causes a compilation failure instead of just removing the corresponding constructor (`layout_stride::mapping` constructor taking a mapping).

See discussion in https://github.com/kokkos/kokkos/pull/7401

There seems to be a secondary and possibly related bug that doesn't let me use `&&` between the two conditions, as GCC doesn't seem to short circuit properly.

### DIFF
--- a/include/experimental/__p0009_bits/layout_stride.hpp
+++ b/include/experimental/__p0009_bits/layout_stride.hpp
@@ -48,9 +48,26 @@ struct layout_right {
 };
 
 namespace detail {
+  // FIXME GCC <= 12: workaround gcc-12 bug that shows up in Kokkos; compilation fails when Mapping doesn't have
+  // extents_type. Normally this should just be a substititon failure, but causes an error with GCC <= 12
+  template<class, class Enabled = void>
+  struct is_mapping_of_impl {
+    template<class>
+    static constexpr bool value_for_layout = false;
+  };
+
+  template<class Mapping>
+  struct is_mapping_of_impl<Mapping, std::void_t<typename Mapping::extents_type>>
+  {
+    // FIXME GCC <= 12: We can't just do a conjunction of the two conditions, because the affected GCC versions seem to not
+    // short-circuit when resolving the substition of Mapping
+    template<class Layout>
+    static constexpr bool value_for_layout = std::is_same<typename Layout::template mapping<typename Mapping::extents_type>, Mapping>::value;
+  };
+
   template<class Layout, class Mapping>
   constexpr bool is_mapping_of =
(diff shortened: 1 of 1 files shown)


include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -48,9 +48,26 @@ struct layout_right {
 };
 
 namespace detail {
+  // FIXME GCC <= 12: workaround gcc-12 bug that shows up in Kokkos; compilation fails when Mapping doesn't have
+  // extents_type. Normally this should just be a substititon failure, but causes an error with GCC <= 12
```
masterleinad:
```suggestion
  // extents_type. Normally this should just be a substitution failure, but causes an error with GCC <= 12
```

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -48,9 +48,26 @@ struct layout_right {
...
+  struct is_mapping_of_impl {
+    template<class>
+    static constexpr bool value_for_layout = false;
+  };
+
+  template<class Mapping>
+  struct is_mapping_of_impl<Mapping, std::void_t<typename Mapping::extents_type>>
+  {
+    // FIXME GCC <= 12: We can't just do a conjunction of the two conditions, because the affected GCC versions seem to not
+    // short-circuit when resolving the substition of Mapping
```
masterleinad:
```suggestion
    // short-circuit when resolving the substitution of Mapping
```

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -48,9 +48,26 @@ struct layout_right {
...
+    template<class Layout>
+    static constexpr bool value_for_layout = std::is_same<typename Layout::template mapping<typename Mapping::extents_type>, Mapping>::value;
```
masterleinad: Do you still need to support C++14 here?

nmm0: Yeah unfortunately... just pushed a fix

include/experimental/__p0009_bits/layout_stride.hpp
nmm0: Thanks apparently I can't type substitution today

masterleinad (approved):

crtrott (approved):

PR #427, opened 2025-09-16
Fix required_span_size for padded layouts with empty extents

### LINKED ISSUE #8460: Span of 0-sized, padded views produces incorrect values
I'm updating my GenTen code to Kokkos 4.7 and am finding weirdness with the span of 0-sized, padded views.  In particular, the following code
```
#include <iostream>
#include "Kokkos_Core.hpp"

int main() {
  Kokkos::initialize();

  {
    Kokkos::View<double**> v(Kokkos::view_alloc("v", Kokkos::AllowPadding),
                             0, 15);
    std::cout << "v.span() = " << v.span() << std::endl;
  }

  Kokkos::finalize();

  return 0;
}
```
produces this output:
```
v.span() = 18446744073709551615
```
which I believe if you cast that to a signed int is -1. With Kokkos 4.6, I get:
```
v.span() = 0
```
which is what I expected.  This is a serial build on MacOS Sequoia 15.6.1 with AppleClang 17.0.0.17000013 compiler.

masterleinad:
Fixes https://github.com/kokkos/kokkos/issues/8460.

### DIFF
--- a/tests/test_layout_padded_left.cpp
+++ b/tests/test_layout_padded_left.cpp
@@ -498,6 +498,19 @@ TEST(LayoutLeftTests, issue362) {
   ASSERT_EQ(mapping.required_span_size(), mapping(1, 1) + 1);
 }
 
+TEST(LayoutLeftTests, empty_span) {
+  {
+    auto mapping = KokkosEx::layout_left_padded<16>::mapping<
+        Kokkos::extents<std::size_t, 0, 15>>();
+    ASSERT_EQ(mapping.required_span_size(), 0);
+  }
+  {
+    auto mapping = KokkosEx::layout_left_padded<16>::mapping<
+        Kokkos::extents<std::size_t, 15, 0>>();
+    ASSERT_EQ(mapping.required_span_size(), 0);
+  }
+}
+
 // https://github.com/kokkos/mdspan/issues/393
 #define LAYOUT_LEFT_COMPILE_ISSUE393_DEATH 0
 TEST(LayoutLeftTests, issue393) {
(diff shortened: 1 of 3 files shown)


nmm0 (approved):

tests/test_layout_padded_right.cpp
```diff
@@ -500,6 +500,19 @@ TEST(LayoutRightTests, issue362) {
   ASSERT_EQ(mapping.required_span_size(), mapping(1, 1) + 1);
 }
 
+TEST(LayoutRightTests, empty_span) {
+  {
+    auto mapping = KokkosEx::layout_left_padded<16>::mapping<
```
science-enthusiast, line 505: Probably, here and at line 510, it should be `layout_right_padded`?

masterleinad: Yes

tests/test_layout_padded_left.cpp
```diff
@@ -498,6 +498,19 @@ TEST(LayoutLeftTests, issue362) {
...
+  {
+    auto mapping = KokkosEx::layout_left_padded<16>::mapping<
+        Kokkos::extents<std::size_t, 0, 15>>();
+    ASSERT_EQ(mapping.required_span_size(), 0);
+  }
+  {
+    auto mapping = KokkosEx::layout_left_padded<16>::mapping<
+        Kokkos::extents<std::size_t, 15, 0>>();
+    ASSERT_EQ(mapping.required_span_size(), 0);
+  }
```
dalg24, line 511: It would be good to have coverage for runtime extent 0 as well

PR #429, opened 2025-10-01
Fix return type of mdspan members

numEricL:
The mdspan members `rank()` and `rank_dynamic()` return `size_t` instead of `extents_type::rank_type` as specified in n4950.


dalg24 (approved):

nmm0 (approved):
Thanks for catching this!

PR #430, opened 2025-10-03
Enable negative strides

numEricL:
I'm writing a pytorch-like wrapper over mdspan. Python-like array slicing allows for negative stride for which there is currently no support for in mdspan. There has been some discussion on this already: #352 
This PR adds some helper functions to compute absolute values for signed types during subspan creation, as well as using std::ptrdiff_t in key areas, such as in the default_accessor.

I also snuck in a default template parameter for strided_slice, otherwise strided_slice{0,3} gives unexpected results.

I've written a quick example that verifies results by printing some small arrays to console. Github is being difficult and won't let me upload .cpp files so I've renamed it as a .txt
[negative_stride.txt](https://github.com/user-attachments/files/22674238/negative_stride.txt)

### DIFF
--- a/include/experimental/__p0009_bits/default_accessor.hpp
+++ b/include/experimental/__p0009_bits/default_accessor.hpp
@@ -42,12 +42,12 @@ struct default_accessor {
 
   MDSPAN_INLINE_FUNCTION
   constexpr data_handle_type
-  offset(data_handle_type p, size_t i) const noexcept {
+  offset(data_handle_type p, ptrdiff_t i) const noexcept {
     return p + i;
   }
 
   MDSPAN_FORCE_INLINE_FUNCTION
-  constexpr reference access(data_handle_type p, size_t i) const noexcept {
+  constexpr reference access(data_handle_type p, ptrdiff_t i) const noexcept {
     return p[i];
   }
 
--- a/include/experimental/__p2630_bits/strided_slice.hpp
+++ b/include/experimental/__p2630_bits/strided_slice.hpp
@@ -30,7 +30,7 @@ namespace detail {
 }
 
 // Slice Specifier allowing for strides and compile time extent
-template <class OffsetType, class ExtentType, class StrideType>
+template <class OffsetType, class ExtentType, class StrideType = std::integral_constant<size_t,1>>
 struct strided_slice {
   using offset_type = OffsetType;
   using extent_type = ExtentType;
--- a/include/experimental/__p2630_bits/submdspan_extents.hpp
+++ b/include/experimental/__p2630_bits/submdspan_extents.hpp
@@ -29,7 +29,7 @@ MDSPAN_TEMPLATE_REQUIRES(
   /* requires */ (std::is_unsigned_v<Unsigned>)
 )
 MDSPAN_INLINE_FUNCTION
-constexpr Unsigned abs(const Unsigned &val) {
+constexpr auto abs(const Unsigned &val) {
   return val;
 }
 
@@ -38,8 +38,8 @@ MDSPAN_TEMPLATE_REQUIRES(
   /* requires */ (std::is_unsigned_v<Unsigned>)
 )
 MDSPAN_INLINE_FUNCTION
-constexpr Unsigned abs(const std::integral_constant<Unsigned, val>&) {
-  return val;
+constexpr auto abs(const std::integral_constant<Unsigned, val>&) {
+  return std::integral_constant<Unsigned, val>();
 }
 
 MDSPAN_TEMPLATE_REQUIRES(
@@ -47,7 +47,7 @@ MDSPAN_TEMPLATE_REQUIRES(
   /* requires */ (std::is_signed_v<Signed>)
 )
 MDSPAN_INLINE_FUNCTION
-constexpr Signed abs(const Signed &val) {
+constexpr auto abs(const Signed &val) {
   return val < 0 ? -val : val;
 }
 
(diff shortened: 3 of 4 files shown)


numEricL:
I probably should have had this as a "draft PR" until now, but I think it's ready for review. Is there a hook to run benchmarks?

mhoemmen:
@numEricL Thank you for your interest in mdspan!  The reference mdspan implementation aims for conformance with the C++ Standard.  The changes proposed in this PR would make the implementation not conform to C++23.

It may be possible to support negative strides with a custom layout mapping, as long as the resulting `required_span_size()` is still nonnegative.  Users can write their own custom layout mappings without changing the mdspan implementation.  For this reason, I am closing this PR.  If you would like to propose adding a new layout mapping as an example instead of making it part of the standard headers, please explain here.

include/experimental/__p0009_bits/default_accessor.hpp
```diff
@@ -42,12 +42,12 @@ struct default_accessor {
 
   MDSPAN_INLINE_FUNCTION
   constexpr data_handle_type
-  offset(data_handle_type p, size_t i) const noexcept {
+  offset(data_handle_type p, ptrdiff_t i) const noexcept {
```
mhoemmen, line 45: This does not conform with the C++ Standard, because `default_accessor` always takes the offset (in both `access` and `offset`) as `size_t`.  Offsets must always be nonnegative.

https://eel.is/c++draft/mdspan.accessor.default.overview

include/experimental/__p2630_bits/strided_slice.hpp
```diff
@@ -30,7 +30,7 @@ namespace detail {
 }
 
 // Slice Specifier allowing for strides and compile time extent
-template <class OffsetType, class ExtentType, class StrideType>
+template <class OffsetType, class ExtentType, class StrideType = std::integral_constant<size_t,1>>
```
mhoemmen, line 33: This does not conform with the current working draft (for C++26), as `strided_slice` has no default template arguments.

https://eel.is/c++draft/mdspan.sub.strided.slice

Please contact your National Body regarding the potential for submitting a comment to change C++26.  The US National Body has already submitted its comments.

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -24,6 +24,59 @@
...
+constexpr auto abs(const Unsigned &val) {
+  return val;
+}
+
+MDSPAN_TEMPLATE_REQUIRES(
+  class Unsigned, Unsigned val,
+  /* requires */ (std::is_unsigned_v<Unsigned>)
+)
+MDSPAN_INLINE_FUNCTION
+constexpr auto abs(const std::integral_constant<Unsigned, val>&) {
```
mhoemmen, line 41: In general, this design would need to work for any _`integral-constant-like`_ type, not just `std::integral_constant`.  Also, [P3663R3](https://isocpp.org/files/papers/P3663R3.html) just passed wording review today and may change C++26, depending on the results of the C++ Standard Committee plenary meeting next month.  That may affect this design.

include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -54,37 +107,72 @@ template <class OffsetType, class ExtentType, class StrideType>
 struct is_strided_slice<
     strided_slice<OffsetType, ExtentType, StrideType>> : std::true_type {};
 
-// Helper for identifying valid pair like things
-template <class T, class IndexType> struct index_pair_like : std::false_type {};
+// Traits for pair like things
```
mhoemmen, line 110: [P3663R3](https://isocpp.org/files/papers/P3663R3.html) just passed wording review today and may affect C++26, depending on the results of the C++ Standard Committee plenary meeting next month.  If P3663 is adopted, this part of the implementation would almost certainly require changes.

numEricL:
@mhoemmen thank you for the feedback!

Does the C++23 spec outright disallow negative strides for mappings or does it allow for unspecified or implementation defined behavior?

The difficulty with negative stride is less so with mappings and more so with allowing pair slices{a, b} with a>b and strided slices with negative stride, however, I see there is similar discussion to this in P3663R3.

mhoemmen:
@numEricL gladly! : - )

> Does the C++23 spec outright disallow negative strides for mappings or does it allow for unspecified or implementation defined behavior?

My reading of [[mdspan.layout.reqmts] 18](https://eel.is/c++draft/views.multidim#mdspan.layout.reqmts-18) is that a stride can only be negative if its corresponding extent is zero, because otherwise some multidimensional index `i...` would exist such that `m(i...)` is negative (thereby violating [[mdspan.layout.reqmts] 8](https://eel.is/c++draft/views.multidim#mdspan.layout.reqmts-8)).

> ...allowing pair slices{a, b} with a>b and strided slices with negative stride....

Would the intent be to support Python-style wrap-around indexing (where negative indices count back from the end)?

numEricL:
@mhoemmen 

Yes I see, it is unambiguous that layout mappings shall not return a negative index. That is unfortunate!

>>    ...allowing pair slices{a, b} with a>b and strided slices with negative stride....

> Would the intent be to support Python-style wrap-around indexing (where negative indices count back from the end)?

I made a mistake in assuming python allowed for reverse ordered pairs, stride must be explicitly prescribed for it to be legal python. I don't support wrap-around indexing, that would remove the possibility of static extents of subspans when the parent span's extents aren't known at compile time. Negative strides are far more permissive, usually known at compile time, and cost-free if one is willing to stick to unsigned types for the stride in strided_slice. Even for signed types it just adds a single absolute value paid for during subspan creation.

mhoemmen:
@numEricL wrote:
> Yes I see, it is unambiguous that layout mappings shall not return a negative index. That is unfortunate!

I don't think many of the mdspan coauthors would agree that it's unfortunate.  It might help to understand what exactly a negative layout mapping result would mean.  Note that it's a key part of the design to prefer that the layout mapping and accessor are independent.

PR #443, opened 2025-12-16
use cuda/std/limits with cuda version >= 12.6

ndellingwood:
this works around compiler issues showing up with downstream usage of kokkos, such as in Trilinos, when testing with cuda/12.4 of the type

```
error: 'iterator_traits' is not a member of 'cuda::std::__4';
```

### DIFF
--- a/include/experimental/__p0009_bits/utility.hpp
+++ b/include/experimental/__p0009_bits/utility.hpp
@@ -4,7 +4,7 @@
 #include <type_traits>
 #include <array>
 #include <utility>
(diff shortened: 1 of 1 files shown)


ndellingwood:
Cross-referencing https://github.com/kokkos/kokkos/pull/8750#issuecomment-3661412495

ndellingwood:
@crtrott @masterleinad for consideration to workaround issues showing up in nightly integration testing of Kokkos with Trilinos with cuda/12.4

dalg24 (approved):

PR #444, opened 2026-01-08
Fix compiling with Cuda 13.1

masterleinad:
Corresponds to https://github.com/kokkos/kokkos/pull/8801.
Thanks @mhoemmen for forwarding the issue and getting the fix for the reproducer!

### DIFF
--- a/include/experimental/__p0009_bits/layout_stride.hpp
+++ b/include/experimental/__p0009_bits/layout_stride.hpp
@@ -221,7 +221,7 @@ struct layout_stride {
 
       MDSPAN_TEMPLATE_REQUIRES(
         class IntegralType,
-        (std::is_convertible<IntegralType, typename extents_type::index_type>::value)
+        (std::is_convertible_v<IntegralType, typename extents_type::index_type>)
       )
       MDSPAN_INLINE_FUNCTION
       // Need to avoid zero length c-array
--- a/include/experimental/__p0009_bits/layout_stride.hpp
+++ b/include/experimental/__p0009_bits/layout_stride.hpp
@@ -221,7 +221,7 @@ struct layout_stride {
 
       MDSPAN_TEMPLATE_REQUIRES(
         class IntegralType,
-        (MDSPAN_IMPL_TRAIT(std::is_convertible_v, IntegralType, typename extents_type::index_type))
+        (MDSPAN_IMPL_TRAIT(std::is_convertible, IntegralType, typename extents_type::index_type))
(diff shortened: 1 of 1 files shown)


mhoemmen (approved):
Thanks for working on this! : - )  Great work with the fix!

For the future (not necessarily this PR), should we consider using `MDSPAN_IMPL_TRAIT` throughout instead of the `::value` or `_v` trait variants explicitly?  `include/experimental/__p0009_bits/macros.hpp` defines `MDSPAN_IMPL_TRAIT` as follows.

```c++
#if MDSPAN_IMPL_USE_VARIABLE_TEMPLATES
#  define MDSPAN_IMPL_TRAIT(TRAIT, ...) TRAIT##_v<__VA_ARGS__>
#else
#  define MDSPAN_IMPL_TRAIT(TRAIT, ...) TRAIT<__VA_ARGS__>::value
#endif
```

Using `MDSPAN_IMPL_TRAIT` thus should make back-porting to C++14 possible, while still requiring use of the `_v` traits in Kokkos (that requires C++17 at least).

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -221,7 +221,7 @@ struct layout_stride {
 
       MDSPAN_TEMPLATE_REQUIRES(
         class IntegralType,
-        (std::is_convertible<IntegralType, typename extents_type::index_type>::value)
+        (MDSPAN_IMPL_TRAIT(std::is_convertible_v, IntegralType, typename extents_type::index_type))
```
mhoemmen: `MDSPAN_IMPL_TRAIT` adds the `_v` for you.

```suggestion
        (MDSPAN_IMPL_TRAIT(std::is_convertible, IntegralType, typename extents_type::index_type))
```

mhoemmen (changes requested):

mhoemmen (approved):
Thanks for the fix!

crtrott (approved):

JBludau:
Could you guys explain a bit what is the problem and how you got to the solution? In the current state I have problems following

masterleinad:
> Could you guys explain a bit what is the problem and how you got to the solution? In the current state I have problems following

We reported the issue to NVIDIA and they found that this change fixes the reproducer. It turns out that it's also enough to fix all compilation problems. Accrding to them, it's an EDG-related bug. There is no good explanation otherwise why this would make any difference.

PR #445, opened 2026-01-19
Fix layout stride converting explicit clause

nmm0:
We forgot to add the padded layouts to the `layout_stride` converting constructor, see https://eel.is/c++draft/mdspan.layout#stride.cons-9

This PR adds tests for all the strided layout conversions/constructions/assignments from other layouts

### DIFF
--- a/compilation_tests/ctest_layout_convertible.cpp
+++ b/compilation_tests/ctest_layout_convertible.cpp
@@ -91,4 +91,22 @@ MDSPAN_STATIC_TEST(
   !std::is_constructible<LS1, NotARealLayout::mapping<E2>>::value
 );
 
+MDSPAN_STATIC_TEST(std::is_constructible<Kokkos::layout_left::mapping<E1>, LS1>::value);
+MDSPAN_STATIC_TEST(std::is_convertible<Kokkos::layout_left::mapping<E1>, LS1>::value);
+MDSPAN_STATIC_TEST(std::is_assignable<LS1, Kokkos::layout_left::mapping<E1>>::value);
 
+MDSPAN_STATIC_TEST(std::is_constructible<Kokkos::layout_right::mapping<E1>, LS1>::value);
+MDSPAN_STATIC_TEST(std::is_convertible<Kokkos::layout_right::mapping<E1>, LS1>::value);
+MDSPAN_STATIC_TEST(std::is_assignable<LS1, Kokkos::layout_right::mapping<E1>>::value);
+
+MDSPAN_STATIC_TEST(std::is_constructible<Kokkos::layout_stride::mapping<E1>, LS1>::value);
+MDSPAN_STATIC_TEST(std::is_convertible<Kokkos::layout_stride::mapping<E1>, LS1>::value);
+MDSPAN_STATIC_TEST(std::is_assignable<LS1, Kokkos::layout_stride::mapping<E1>>::value);
+
+MDSPAN_STATIC_TEST(std::is_constructible<Kokkos::Experimental::layout_left_padded<14>::mapping<E1>, LS1>::value);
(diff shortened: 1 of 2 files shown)


include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -441,7 +442,9 @@ struct layout_stride {
       !(std::is_convertible<typename StridedLayoutMapping::extents_type, extents_type>::value &&
        (detail::is_mapping_of<layout_left, StridedLayoutMapping> ||
         detail::is_mapping_of<layout_right, StridedLayoutMapping> ||
-        detail::is_mapping_of<layout_stride, StridedLayoutMapping>))
+        detail::is_mapping_of<layout_stride, StridedLayoutMapping> ||
+        MDSPAN_IMPL_PROPOSED_NAMESPACE::detail::is_layout_left_padded_mapping<StridedLayoutMapping>::value || // Don't need to guard for C++14 as this isn't compiled in < C++20
+        MDSPAN_IMPL_PROPOSED_NAMESPACE::detail::is_layout_right_padded_mapping<StridedLayoutMapping>::value))
```
mhoemmen:
```suggestion
        MDSPAN_IMPL_PROPOSED_NAMESPACE::detail::is_layout_left_padded_mapping<StridedLayoutMapping>::value || // Don't need to guard for C++14 as this isn't compiled in < C++20
        MDSPAN_IMPL_PROPOSED_NAMESPACE::detail::is_layout_right_padded_mapping<StridedLayoutMapping>::value ||
        detail::is_mapping_of<layout_stride, StridedLayoutMapping>))
```

mhoemmen: Should we consider using the same order as in the Standard wording?  It doesn't matter for correctness, but it is easier to tell that way that the implementation matches the wording.

mhoemmen (approved):
Thanks for the fix!

compilation_tests/ctest_layout_convertible.cpp
```diff
@@ -91,4 +91,24 @@ MDSPAN_STATIC_TEST(
   !std::is_constructible<LS1, NotARealLayout::mapping<E2>>::value
 );
 
-
+MDSPAN_STATIC_TEST(std::is_constructible<Kokkos::layout_left::mapping<E1>, LS1>::value);
+MDSPAN_STATIC_TEST(std::is_convertible<Kokkos::layout_left::mapping<E1>, LS1>::value);
+MDSPAN_STATIC_TEST(std::is_assignable<LS1, Kokkos::layout_left::mapping<E1>>::value);
```
crtrott, line 96: all of these test make not much sense to me the way its written it needs to be:
```c++
is_constructible<To, From>
is_convertible<From, To>
is_assignable<To, From>
```

crtrott: Given that we are talking about going to layout stride in this PR I think it needs to be:
```c++
MDSPAN_STATIC_TEST(std::is_constructible<LS1, Kokkos::layout_left::mapping<E1>>::value);
MDSPAN_STATIC_TEST(std::is_convertible<Kokkos::layout_left::mapping<E1>, LS1>::value);
MDSPAN_STATIC_TEST(std::is_assignable<LS1, Kokkos::layout_left::mapping<E1>>::value);
```

crtrott (changes requested):

include/experimental/__p0009_bits/layout_stride.hpp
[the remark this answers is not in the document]
nmm0: Yeah... this isn't correct.

compilation_tests/ctest_layout_convertible.cpp
nmm0: Woops, yeah I must have done a copy and paste bug

crtrott (approved):

PR #422, opened 2025-06-18
Port `aligned_accessor` from the Raft fork of Kokkos to mainline.

vitor1001:
See https://github.com/rapidsai/raft/blob/branch-25.08/cpp/include/raft/thirdparty/mdspan/include/experimental/__p0009_bits/aligned_accessor.hpp.

This implementation was adapted from kokkos/mdspan/blob/stable/examples/aligned_accessor/aligned_accessor.cpp.

Fixes https://github.com/kokkos/mdspan/issues/412.

PR #446, opened 2026-01-28
add aligned accessor as part of the library API

### LINKED ISSUE #412: Feature request: support aligned_accessor for C++20 code
I just noticed that Raft is using std::experimental::aligned_accessor in its Kokkos fork in its thirdparty/ subdir [[link](https://github.com/rapidsai/raft/blob/8bd5605f53500b7e3f70579c521fdbfcec6439e5/cpp/include/raft/thirdparty/mdspan/include/experimental/__p0009_bits/aligned_accessor.hpp)]. It builds fine on C++20.

It would be very nice to upstream it to kokkos/mdspan mainline.

nmm0:
Rebase/rework of #422

Closes #412


include/experimental/__p0009_bits/aligned_accessor.hpp
```diff
@@ -0,0 +1,209 @@
...
+// so we can check whether the byte alignment is valid.
+// This makes it impossible to use the alias
+// with an invalid byte alignment.
+template<class T, std::size_t Alignment>
+struct aligned_pointer {
+  static_assert(valid_byte_alignment<T>(Alignment),
+		"Alignment must be a power of two no less than "
+		"the minimum required alignment of T.");
+  using type = T* MDSPAN_IMPL_ALIGN_VALUE_ATTRIBUTE( Alignment );
+};
```
crtrott, line 138: I don't think this is necessary is it? If you have a static_assert inside the `aligned_accessor` you can't get to `aligned_accessor::data_handle_type` without this being checked or?

examples/aligned_accessor/aligned_accessor.cpp
```diff
@@ -34,6 +34,10 @@
 // https://github.com/rapidsai/raft/pull/725#discussion_r937991701
 
 namespace {
+using Kokkos::aligned_accessor;
+using Kokkos::detail::aligned_pointer_t;
+using Kokkos::detail::assume_aligned_method;
```
crtrott, line 39: I am not sure I like that we are adding this. These guys are used only in this example, but we are putting them into the main library (`assume_aligned_method` and `align_attribute_method`). I think we should remove them altogether.

tests/test_aligned_accessor.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+  testing::aligned_free(buff);
+}
+
+TEST(TestAlignedAccessor, AlignedAccessor) {
+  test_aligned_accessor<alignof(double), 10>();
+  test_aligned_accessor<alignof(double), 10>();
+  test_aligned_accessor<2 * alignof(double), 10>();
+  test_aligned_accessor<4 * alignof(double), 10>();
+}
```
crtrott, line 93: Can we add a test of something where the sizeof and alignof are different. I.e. use something like:

```c++
struct SomeStruct {
  int val1;
  double val2;
};
```

crtrott (changes requested):
OK I requested some changes.

examples/aligned_accessor/aligned_accessor.cpp
mhoemmen: +1, we do not need `aligned_pointer_t` or `assume_aligned_method` in Standard C++, so we should not expose them to users.

examples/aligned_accessor/aligned_accessor.cpp
```diff
@@ -43,129 +47,13 @@ constexpr std::size_t min_byte_alignment = min_overalignment_factor * sizeof(flo
...
-
-#if defined(__ICC)
-#  pragma warning pop
-#endif
-};
-
-template<class ElementType, std::size_t byte_alignment>
-using aligned_pointer_t = typename aligned_pointer<ElementType, byte_alignment>::type;
-
 template<class ElementType, std::size_t byte_alignment>
```
mhoemmen, line 50: The Standard C++ way to spell this is `std::assume_aligned<byte_alignment>(ptr)`.

include/experimental/__p0009_bits/aligned_accessor.hpp
mhoemmen: The original point of this was that some compiler (I think MSVC?) needed a special attribute marker on pointers to declare overalignment.  That's not the Standard C++ way of doing things, but it might be reasonable for pre-C++20 back-ports.  If you have C++20, the Standard way to do this is to use `assume_aligned`.

include/experimental/__p0009_bits/aligned_accessor.hpp
```diff
@@ -0,0 +1,209 @@
...
+template<class ElementType, std::size_t byte_alignment>
+using aligned_pointer_t = typename aligned_pointer<ElementType, byte_alignment>::type;
+} // namespace detail
+
+template<size_t Alignment, class T>
+#ifdef __cpp_lib_bit_cast // Only can be constexpr if we have bit_cast
+constexpr
+#endif
+bool is_sufficiently_aligned(T* ptr) {
+  // Note this mandate is not what is currently in the standard
```
mhoemmen, line 150: LWG ended up resolving this by Mandating that `Alignment` be a power of two, but _not_ requiring that `Alignment >= alignof(T)`.

include/experimental/__p0009_bits/aligned_accessor.hpp
mhoemmen: Thus, I think this is acceptable, as long as it is not exposed to users.

include/experimental/__p0009_bits/layout_padded.hpp
```diff
@@ -0,0 +1,783 @@
+/*
```
mhoemmen: Should we consider merging the padded layouts separately?  It looks like this might be the result of an incorrect rebase.

mhoemmen (changes requested):

include/experimental/__p0009_bits/layout_padded.hpp
crtrott: Yeah agreed!

include/experimental/__p0009_bits/aligned_accessor.hpp
crtrott: Yeah but we are only using that typedef anyway, so we might as well just declare it exactly this way inside `aligned_accessor`? I.e. `aligned_accessor<T, 4>::data_handle_type` is not `aligned_pointer<T,4>` its `aligned_pointer<T,4>::type` so we can make that typedef inline in `aligned_accessor` or?

examples/aligned_accessor/aligned_accessor.cpp
nmm0: This was mostly to merge Mark's example with the new code. This isn't exposing to the API as they are under `detail::`

include/experimental/__p0009_bits/layout_padded.hpp
nmm0: Not 100% sure what happened here lol

nmm0: Maybe it was introduced in the original PR and I didn't notice, I removed it now.

include/experimental/__p0009_bits/aligned_accessor.hpp
mhoemmen: @crtrott Ah, right, that makes total sense then.  At least `aligned_pointer_t<T, 4>` would be more idiomatic.

examples/aligned_accessor/aligned_accessor.cpp
mhoemmen: It makes sense that the `aligned_accessor` example needs to pull in `aligned_accessor`.  However, why does the example need `aligned_pointer_t` or `assume_aligned_method`?  Wouldn't those just be implementation details of `aligned_accessor` to help with back-porting it to C++ versions earlier than 20?

examples/aligned_accessor/aligned_accessor.cpp
```diff
@@ -43,129 +47,13 @@ constexpr std::size_t min_byte_alignment = min_overalignment_factor * sizeof(flo
...
 aligned_pointer_t<ElementType, byte_alignment>
 bless(ElementType* ptr, std::integral_constant<std::size_t, byte_alignment> /* ba */ )
```
mhoemmen, line 52: Should we consider just calling this `assume_aligned`?  More importantly, why does the example need this?  In general, we should favor the example looking as much like Standard C++ as possible.  (Yes, I know I wrote the example! but it still could use revision : - ) .)

include/experimental/__p0009_bits/aligned_accessor.hpp
```diff
@@ -0,0 +1,209 @@
...
+// NOTE: This code is prematurely taken from an example based on
+// https://github.com/kokkos/mdspan/pull/176
+
+#pragma once
+
+#include "macros.hpp"
+#include "trait_backports.hpp"
+#include "default_accessor.hpp"
+#include "extents.hpp"
+#include <bit>
```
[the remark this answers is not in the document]
mhoemmen: @Copilot That's the right thinking!  That's just not the way to check whether the header exists.  The right way would be first the language version, then the feature test macro, either `__cpp_lib_bitops` or `__cpp_lib_int_pow2`.  Please see the list of feature test macros here: https://en.cppreference.com/w/cpp/utility/feature_test.html .

include/experimental/__p0009_bits/aligned_accessor.hpp
```diff
@@ -0,0 +1,209 @@
...
+#include <cassert>
+#include <iostream>
```
mhoemmen, line 56: Should we consider removing this include?

examples/aligned_accessor/aligned_accessor.cpp
nmm0: Yeah I tried to make as minimum possible changes to the example as possible, but now I'm reconsidering that since, well, the example is supposed to be useful to learn how to use the API. So I might rewrite a lot of it (and possible just make it require C++20)

include/experimental/__p0009_bits/aligned_accessor.hpp
```diff
@@ -0,0 +1,209 @@
...
+#elif defined(__clang__)
+#  define MDSPAN_IMPL_ALIGN_VALUE_ATTRIBUTE( BYTE_ALIGNMENT ) __attribute__((align_value( BYTE_ALIGNMENT )))
+  constexpr char align_attribute_method[] = "__attribute__((align_value(BYTE_ALIGNMENT)))";
+#else
+#  define MDSPAN_IMPL_ALIGN_VALUE_ATTRIBUTE( BYTE_ALIGNMENT )
+  constexpr char align_attribute_method[] = "(none)";
+#endif
+
+constexpr bool
+is_nonzero_power_of_two(const std::size_t x)
```
mhoemmen, line 110: Should we consider just calling this `has_single_bit`?  That would make the implementation look as much like Standard C++ as possible.

tests/test_aligned_accessor.cpp
```diff
@@ -0,0 +1,93 @@
+
+#include <gtest/gtest.h>
+#include <mdspan/mdspan.hpp>
+#include <numeric>
+#include <cstdlib>
+
+TEST(TestAlignedAccessor, IsSufficientlyAligned) {
+  ASSERT_TRUE(
+      Kokkos::is_sufficiently_aligned<1>(reinterpret_cast<char *>(0x12345678)));
```
mhoemmen, line 9: This is technically UB, because we're forming a likely invalid pointer.  Another way we could test alignment of 1 would be to construct a array of `char` aligned to (say) 4 or 8 bytes at least, and then take the address of element 1.

examples/aligned_accessor/aligned_accessor.cpp
mhoemmen: Thanks for considering this!  There's nothing wrong with back-porting, but I think I would want both the example and the `aligned_accessor` implementation to look as much like Standard C++ as possible.  The latter could be achieved just by changing some names.

nmm0: Alright, will do then, I should have a new version tomorrow with all the changes

PR #447, opened 2026-02-02
move padded layouts and dims from Experimental to the standard namespace

### LINKED ISSUE #404: move padded layout out of experimental
Padded layouts are merged into the C++26 standard, we don't need them in experimental anymore

nmm0:
Closes #404

### DIFF
--- a/include/experimental/__p0009_bits/layout_stride.hpp
+++ b/include/experimental/__p0009_bits/layout_stride.hpp
@@ -445,8 +445,8 @@ struct layout_stride {
       !(std::is_convertible<typename StridedLayoutMapping::extents_type, extents_type>::value &&
        (detail::is_mapping_of<layout_left, StridedLayoutMapping> ||
         detail::is_mapping_of<layout_right, StridedLayoutMapping> ||
-        MDSPAN_IMPL_PROPOSED_NAMESPACE::detail::is_layout_left_padded_mapping<StridedLayoutMapping>::value || // Don't need to guard for C++14 as this isn't compiled in < C++20
-        MDSPAN_IMPL_PROPOSED_NAMESPACE::detail::is_layout_right_padded_mapping<StridedLayoutMapping>::value ||
+        detail::is_layout_left_padded_mapping<StridedLayoutMapping>::value || // Don't need to guard for C++14 as this isn't compiled in < C++20
+        detail::is_layout_right_padded_mapping<StridedLayoutMapping>::value ||
         detail::is_mapping_of<layout_stride, StridedLayoutMapping>))
     ) // needs two () due to comma
     MDSPAN_INLINE_FUNCTION MDSPAN_IMPL_CONSTEXPR_14
--- a/benchmarks/fill.hpp
+++ b/benchmarks/fill.hpp
@@ -18,8 +18,6 @@
 
 #include <mdspan/mdspan.hpp>
 
-namespace KokkosEx = MDSPAN_IMPL_STANDARD_NAMESPACE::MDSPAN_IMPL_PROPOSED_NAMESPACE;
-
 #include <benchmark/benchmark.h>
 
 #include <memory>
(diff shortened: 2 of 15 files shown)


crtrott (approved):

mhoemmen (changes requested):
Should we consider also regenerating the single-header file for Compiler Explorer as part of this PR?  It would be excellent for the single-header version to have the same namespaces as the installable version.  Thanks! : - )

crtrott:
> Should we consider also regenerating the single-header file for Compiler Explorer as part of this PR? It would be excellent for the single-header version to have the same namespaces as the installable version. Thanks! : - )

Let's do that in a separate PR, and maybe after your big guy is through. We also are trying to actually create a release, likely coinciding with Kokkos 5.1.0. That opens an interesting question though: currently Kokkos leadership is leaning towards a recommendation (not enforcement) that we have a unified versioning. The idea being that it will be less confusing inside the Kokkos ecosystem about what belongs together and is tested together. I.e. some of us are just exasperated by stuff like the Visual Studio / MSVC versioning, or even the CUDA versioning where it's kinda hard to decide what is what. Any strong feelings on that? If we go with that recommendation for mdspan, we would release as mdspan 5.1 ...

mhoemmen:
@mhoemmen wrote:
> Should we consider also regenerating the single-header file for Compiler Explorer as part of this PR?

@crtrott wrote:
> Let's do that in a separate PR, and maybe after your big guy is through.

That's fine!  PR https://github.com/kokkos/mdspan/pull/408 actually fixes the `dims` namespace issue (implementing acceptance of P2389 in C++26) as a drive-by, because it was easier to use `dims` to write tests.  It also adds a test for `dims`.

@crtrott wrote:
> ... currently Kokkos leadership is leaning towards a recommendation (not enforcement) that we have a unified versioning.... Any strong feelings on that? If we go with that recommendation for mdspan, we would release as mdspan 5.1 ...

I think that's a reasonable idea.  I'm talking it over with colleagues.  Would you plan the same thing for the reference std::linalg implementation as well?

include/experimental/__p0009_bits/layout_stride.hpp
```diff
@@ -445,8 +445,8 @@ struct layout_stride {
       !(std::is_convertible<typename StridedLayoutMapping::extents_type, extents_type>::value &&
        (detail::is_mapping_of<layout_left, StridedLayoutMapping> ||
         detail::is_mapping_of<layout_right, StridedLayoutMapping> ||
-        MDSPAN_IMPL_PROPOSED_NAMESPACE::detail::is_layout_left_padded_mapping<StridedLayoutMapping>::value || // Don't need to guard for C++14 as this isn't compiled in < C++20
-        MDSPAN_IMPL_PROPOSED_NAMESPACE::detail::is_layout_right_padded_mapping<StridedLayoutMapping>::value ||
+        detail::is_layout_left_padded_mapping<StridedLayoutMapping>::value || // Don't need to guard for C++14 as this isn't compiled in < C++20
```
mhoemmen, line 448: 1. It's not clear how this particular line is relevant to C++14 vs. C++20.  Of course `MDSPAN_CONDITIONAL_EXPLICIT` would make this code go away for < 17.

2. Don't we require C++17 as well?  The PR check-in tests currently exercise 14, 17, 20, and 23.

mhoemmen (approved):

include/experimental/__p0009_bits/layout_stride.hpp
nmm0: Yeah the main thing is that padded layout mappings requires C++17 (doesn't work with C++14) -- so normally we should not be directly be using it in unguarded code but as you pointed out `MDSPAN_CONDITIONAL_EXPLICIT` guards it already behind C++20

nmm0: the comment was to make it clear it wasn't a bug to omit checks for C++17

mhoemmen: Thanks for explaining! : - )

PR #457, opened 2026-06-26
Use canonicalize slices in submdspan

crtrott:
This is the last piece to be extracted from PR #408 to implement P3663. After this we need to introduce `extent_slice`.


include/experimental/__p0009_bits/config.hpp
```diff
@@ -45,11 +45,13 @@
 #else
 #define MDSPAN_CXX_STD_23 202100L
 #endif
+#define MDSPAN_CXX_STD_26 202603L
 
 #define MDSPAN_HAS_CXX_14 (MDSPAN_IMPL_CPLUSPLUS >= MDSPAN_CXX_STD_14)
 #define MDSPAN_HAS_CXX_17 (MDSPAN_IMPL_CPLUSPLUS >= MDSPAN_CXX_STD_17)
 #define MDSPAN_HAS_CXX_20 (MDSPAN_IMPL_CPLUSPLUS >= MDSPAN_CXX_STD_20)
 #define MDSPAN_HAS_CXX_23 (MDSPAN_IMPL_CPLUSPLUS >= MDSPAN_CXX_STD_23)
+#define MDSPAN_HAS_CXX_26 (MDSPAN_IMPL_CPLUSPLUS >= MDSPAN_CXX_STD_26)
```
mhoemmen, line 54: Given that this is for checking whether `std::complex` is _`tuple-like`_, should we consider having a finer-grained check?  The feature test macro check would look like this,

```
(__cpp_lib_tuple_like >= 202311L)
```

as P2819 updated the value from `202207L`.

mhoemmen: On the other hand, we might later want to use this macro for other C++26 features like parameter pack indexing.

include/experimental/__p2630_bits/submdspan.hpp
```diff
@@ -27,7 +27,19 @@ MDSPAN_INLINE_FUNCTION
 constexpr auto
 submdspan(const mdspan<ElementType, Extents, LayoutPolicy, AccessorPolicy> &src,
           SliceSpecifiers... slices) {
-  const auto sub_submdspan_mapping_result = submdspan_mapping(src.mapping(), slices...);
+
+  // Avoid instantiating expensive slice mandate check here for known layouts
```
mhoemmen: Are the known layouts just calling this function anyway?  If so, then it might not be much more expensive to do it twice, since you've already instantiated the function once.

mhoemmen (approved):

include/experimental/__p2630_bits/submdspan.hpp
crtrott: Not sure if it actually caches that properly ...

mhoemmen: @crtrott What I'm getting at is that we're diverging from the plain wording of the Standard for the sake of an unproven optimization.  It adds a compile-time check, so it's not an obvious win.

include/experimental/__p0009_bits/config.hpp
crtrott: Yeah we could test more finely for the tuple-like I guess? But we can have the CXX26 thing anyway.

nmm0 (approved):
I guess we have to put in breaking changes when this gets snapshotted into Kokkos that you can't use complex as slice specifiers without C++26? IDK if it matters that much it seems unusual

crtrott:
We don't support complex as slice specifier to subview anyway.

PR #459, opened 2026-07-01
Remove templated get function from static_array_impl::get

Quntized:
While reviewing the mdspan. I noticed a missing return statement in the C++14 implementation of `static_array_impl::get<r>()` at `include/experimental/__p0009_bits/extents.hpp` .

### DIFF
--- a/include/experimental/__p0009_bits/extents.hpp
+++ b/include/experimental/__p0009_bits/extents.hpp
@@ -94,7 +94,7 @@ struct static_array_impl<R, T, FirstExt, Extents...> {
     else
       return static_array_impl<R + 1, T, Extents...>::template get<r>();
 #else
-    get(r);
+    return get(r);
 #endif
   }
 };
--- a/include/experimental/__p0009_bits/extents.hpp
+++ b/include/experimental/__p0009_bits/extents.hpp
@@ -87,16 +87,6 @@ struct static_array_impl<R, T, FirstExt, Extents...> {
     else
       return static_array_impl<R + 1, T, Extents...>::get(r);
   }
-  template <size_t r> MDSPAN_INLINE_FUNCTION constexpr static T get() {
-#if MDSPAN_HAS_CXX_17
-    if constexpr (r == R)
-      return FirstExt;
-    else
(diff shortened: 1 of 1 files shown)


mhoemmen:
Thanks for the fix!

Could you please explain your interest in C++14 support?  Do you have particular regulatory requirements to use C++14 or did you just happen to notice the bug?  Also, did you use an automated tool to find this, and if so, what tool?

I'm asking because much of the value of mdspan comes from newer C++ versions.  The `submdspan` feature (for array slices) requires at least C++17, for example.

Quntized:
Hi @mhoemmen, thanks for the reply!

I don't have any regulatory requirements for C++14, nor did I use any automated tools. I am just exploring the Kokkos codebase to better understand its architecture and template metaprogramming patterns. While manually reading and testing through the mdspan implementations, I spotted the missing return statement in the C++14 #else block.

Interestingly, when I tried to trigger the bug locally on my laptop by forcing CMAKE_CXX_STANDARD=14, it still bypassed the fallback path. I realized CMake was applying transitive compiler requirements-since my local Kokkos installation was built with C++20, causing MDSPAN_HAS_CXX_17 to evaluate to true. However, I knew a strict C++14 setup compiling from scratch would still hit the missing return.

I know the library is focused on modern C++ now, but I figured it was worth submitting the patch for anyone still locked into older environments.

If it's useful context , I maintain a personal repository where I systematically test and investigate Kokkos internals while exploring the codebase: https://github.com/Quntized/Kokkos_internal. The C++14 path in mdspan came up while doing similar manual exploration of the containers layer

crtrott:
Interesting out C++ tests here in mdspan must not exercise that guy. I think we can just merge it for now. We may wanna try and figure out why it's not getting exercised though.

crtrott (approved):

crtrott:
I think the entire function is unused at this point and could just be deleted.

crtrott (changes requested):
Lets just remove the get function which is templated on the index value.

crtrott (approved):

nmm0 (approved):
Thanks for the fix!

PR #462, opened 2026-07-16
Silence SYCL unused parameter warning

NamjaeChoi:


masterleinad (commented):
Since `mdspan` tries to be C++14 compliant in part, it might be better to do
```C++
ifdef __SYCL_DEVICE_ONLY__
  (void) a;
  (void) b;
  return true;
#else
```
instead.

masterleinad (approved):

mhoemmen (approved):
Thanks for the contribution!

The changes look harmless, though I have no way to test them currently.

PR #463, opened 2026-07-23
Add an OpenACC variant of default_precondition_violation_handler()

seyonglee:
This PR adds an OpenACC variant of mdspan default_precondition_violation_handler() function, which fixes the unsupported fprintf error in Kokkos PR #9345 (https://github.com/kokkos/kokkos/pull/9345#issuecomment-5046076022).
This PR contains the same changes as Kokkos PR #9369 (https://github.com/kokkos/kokkos/pull/9369).

### DIFF
--- a/include/experimental/__p0009_bits/macros.hpp
+++ b/include/experimental/__p0009_bits/macros.hpp
@@ -24,7 +24,7 @@
 #if defined(MDSPAN_IMPL_HAS_SYCL)
 #include <sycl/sycl.hpp> // sycl::ext::oneapi::experimental::printf
 #endif
-#if defined(MDSPAN_IMPL_HAS_CUDA) || defined(MDSPAN_IMPL_HAS_HIP) || defined(MDSPAN_IMPL_HAS_SYCL)
+#if defined(MDSPAN_IMPL_HAS_CUDA) || defined(MDSPAN_IMPL_HAS_HIP) || defined(MDSPAN_IMPL_HAS_SYCL) || defined(MDSPAN_IMPL_HAS_OPENACC)
 #include "assert.h"
 #endif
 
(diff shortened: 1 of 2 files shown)


nmm0 (approved):

include/experimental/__p0009_bits/macros.hpp
```diff
@@ -130,6 +130,14 @@ MDSPAN_FUNCTION inline void default_precondition_violation_handler(const char* c
...
   assert(0);
 }
+#elif defined(MDSPAN_IMPL_HAS_OPENACC)
+MDSPAN_FUNCTION inline void default_precondition_violation_handler(const char* cond, const char* file, unsigned line)
+{
+  (void) cond;
+  (void) file;
+  (void) line;
+  assert(0);
+}
```
mhoemmen, line 137:
```suggestion
#elif defined(MDSPAN_IMPL_HAS_OPENACC)
MDSPAN_FUNCTION inline void default_precondition_violation_handler(const char* /* cond */, const char* /* file */, unsigned /* line */)
{
  assert(false);
}
```

mhoemmen (changes requested):
Would you consider a small change, just to remove the possibility of unused parameter warnings completely?  Thanks! : - )

include/experimental/__p0009_bits/macros.hpp
mhoemmen, line 142: This is a separate issue, but should we consider putting `assert(false);` here instead of `std::abort();`?  It doesn't seem right to make this back end abort unconditionally, but have all the other back-ends use the `assert` macro.

seyonglee:
> Would you consider a small change, just to remove the possibility of unused parameter warnings completely? Thanks! : - )

Applied; thanks.

mhoemmen (approved):
Thanks for the fix!

include/experimental/__p0009_bits/macros.hpp
dalg24: Why would we want the default precondition handler to silently do nothing when NDEBUG is defined?

dalg24: Hmmm I see now that is what we do for CUDA/HIP/SYCL so we are inconsistent in the fallback implementation that you pointed out.

mhoemmen: @dalg24 That's right!  Whatever we do, we should be consistent about it.

PR #467, opened 2026-09-02
Change global tag variables to local variables.

seyonglee:
This PR changes global tag variables accessed in Kokkos subview-related functions to local variables, so that NVHPC OpenACC compiler (nvc++ version higher than 24.5) can compile a Kokkos program where a subview is created in a device kernel.
This PR contains the same changes as mdspan code changes in Kokkos PR 9486 (https://github.com/kokkos/kokkos/pull/9486)

### DIFF
--- a/include/experimental/__p0009_bits/layout_stride.hpp
+++ b/include/experimental/__p0009_bits/layout_stride.hpp
@@ -367,7 +367,7 @@ struct layout_stride {
 #else
       : base_t(base_t{member_pair_t(
 #endif
-          e, strides_storage_t(deduction_workaround_impl::fill_strides(mdspan_non_standard, s))
+          e, strides_storage_t(deduction_workaround_impl::fill_strides(mdspan_non_standard_tag(), s))
 #if defined(MDSPAN_IMPL_USE_ATTRIBUTE_NO_UNIQUE_ADDRESS)
         }
 #else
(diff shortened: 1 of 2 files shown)


nmm0:
Do you know why this is causing problems? Is it an nvc++ compiler bug?

seyonglee:
> Do you know why this is causing problems? Is it an nvc++ compiler bug?

If a subview is created in a device kernel (e.g., in a Kokkos parallel_for()), several global tag symbols are accessed, which cannot be correctly captured by nvc++ compiler if version > 24.5.

nmm0:
Fascinating, even if they are constexpr? Well either way I think this is fine

nmm0 (approved):

dalg24 (commented):
Just noting that `mdspan_non_standard` would now become unused.

cc @mhoemmen @weinbe2

mhoemmen:
Thanks for the fix! : - ) 

@seyonglee wrote:
> ... so that NVHPC OpenACC compiler (nvc++ version higher than 24.5) can compile a Kokkos program where a subview is created in a device kernel.

Does this mean "is able to build without errors" or "is able to build and pass tests"?

seyonglee:
> Thanks for the fix! : - )
> 
> @seyonglee wrote:
> 
> > ... so that NVHPC OpenACC compiler (nvc++ version higher than 24.5) can compile a Kokkos program where a subview is created in a device kernel.
> 
> Does this mean "is able to build without errors" or "is able to build and pass tests"?

With this update, the following two Kokkos unit tests, which create subview in a device kernel, can be built without errors and pass tests:

core/unit_test/range_policy/TestRangePolicyExecutionTypes.hpp
core/unit_test/TestSubView_c13.hpp

mhoemmen:
@seyonglee I'm just seeing what you wrote now -- thanks for explaining how you reproduced the issue!  I wouldn't object to someone merging this PR, but I'd like to take a closer look at this.

PR #469, opened 2026-09-09
Raise CMake minimum required version to 3.22

dalg24:
To align with Kokkos (Core) and to avoid having to set a number of policies when embedding in Kokkos.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.22)
 
 project(MDSpan
   VERSION 0.6.0
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -49,8 +49,6 @@ else()
       GIT_TAG v1.17.0
   )
   FetchContent_MakeAvailable(googletest)
-  
-  add_library(GTest::gtest_main ALIAS gtest_main)
 endif()
 
 # FIXME_MSVC
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -43,6 +43,7 @@ else()
   if (MSVC)
     set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
   endif()
+  set(INSTALL_GTEST OFF)
   FetchContent_Declare(
       googletest
       GIT_REPOSITORY https://github.com/google/googletest.git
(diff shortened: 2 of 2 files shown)


crtrott (approved):

crtrott:
Of course we need to fix the CI breakage.

tests/CMakeLists.txt
```diff
@@ -43,20 +43,12 @@ else()
...
       GIT_REPOSITORY https://github.com/google/googletest.git
       GIT_TAG v1.17.0
   )
-  
-  # TODO CMake 3.28, we can pass EXCLUDE_FROM_ALL directly to fetchcontent_makeavailable
-  fetchcontent_getproperties(googletest)
-  if (NOT googletest_POPULATED)
-    fetchcontent_populate(googletest)
-    add_subdirectory(${googletest_SOURCE_DIR} ${googletest_BINARY_DIR} EXCLUDE_FROM_ALL)
-  endif()
```
dalg24, line 57: I am not sure what this was trying to achieve?
Not build gmock and gmock_main?

masterleinad: Looks like th depreacted pattern described in https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_getproperties.

nmm0: We have to use the old pattern here since `EXCLUDE_FROM_ALL` is not supported by `FetchContent_MakeAvailable` until CMake 3.28

nmm0 (changes requested):
We need to make sure google test is not installed when mdspan is installed, see commit https://github.com/kokkos/mdspan/commit/eb2989d40d30a341f137143abdd2032d27f3834b

masterleinad (approved):

nmm0 (approved):

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.22)
```
dalg24, line 1: @mhoemmen do you have any objection to this change?

mhoemmen:
@dalg24 HPC SDK team confirms that this is OK.  Thanks for asking!

CMakeLists.txt
mhoemmen: @dalg24 Neither the HPC SDK team nor the CCCL team have objections to this change.  Thanks for asking! : - )

